PR #112, opened 2017-11-13
Italy special

pysailor:
Exceptions to the client software / behaviour that are necessary for the way OIRA is used in Italy

### DIFF
--- a/src/euphorie/client/risk.py
+++ b/src/euphorie/client/risk.py
@@ -12,6 +12,7 @@ from Acquisition import aq_inner
 from euphorie.client import model
 from euphorie.client.interfaces import IActionPlanPhaseSkinLayer
 from euphorie.client.interfaces import IIdentificationPhaseSkinLayer
+from euphorie.client.interfaces import IItalyIdentificationPhaseSkinLayer
 from euphorie.client.navigation import FindNextQuestion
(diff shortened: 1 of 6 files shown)


src/euphorie/client/risk.py
```diff
@@ -161,6 +162,11 @@ def update(self):
                 custom_ds = getattr(self.request.survey, 'description_severity', '') or ''
                 self.description_severity = custom_ds.strip() or self.description_severity
 
+            # Italian special
+            if IItalyIdentificationPhaseSkinLayer.providedBy(self.request):
+                self.skip_evaluation = True
```
ale-rt: lol ;)

src/euphorie/client/utils.py
```diff
@@ -219,6 +220,20 @@ def logoMode(self):
...
+    def styles_override(self):
+        css = ""
+        if IItaly.providedBy(self.request):
+            css = """
+#steps .topics .legend li.answered.risk::before {
+    background: purple;
+}
+#steps .topics .questions li.answered.risk a::before {
+    background: purple;
+}
```
wichert: Why not add the country to the body as a class (like `body class="country-it"`) and use that in the standard CSS?

pysailor: Yes, would have been an option too

PR #121, opened 2018-04-04
Fix undefined variable

### LINKED ISSUE #120: local variable 'show_problem_description' referenced before assignment
Under some circumstances the variable show_problem_description is undefined:
```
2018-04-04 17:39:49 ERROR euphorie.client.error Error at <euphorie.ghost.PathGhost object at 0x7f98addbce90>
...
    if has_risk and not show_problem_description:
UnboundLocalError: local variable 'show_problem_description' referenced before assignment
```

See: https://github.com/euphorie/Euphorie/blob/f2dac338ee49f0d2a32711df0ea45c78d327ae17/src/euphorie/client/report.py#L544

ale-rt:
Fixes #120

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,6 +5,8 @@ Changelog
 -------------------
 
 - Translation updates in Castillian (es) and Catalan (ca)
+- Fix undefined variable (#120)
+
 
 9.0.35 (2018-03-16)
 -------------------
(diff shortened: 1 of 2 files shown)


ale-rt:
In this PR I am setting the default for `show_problem_description` to `False`.
I am not really if it is a sane default.

PR #129, opened 2018-05-04
Display sessions as a tree

ale-rt:

PR #131, opened 2018-05-04
Clean up

ale-rt:
Apparently some methods are not needed anymore


pysailor:
Does this replace #129?

ale-rt:
That is build on top of it, unluckily #129 has failing tests.

PR #132, opened 2018-05-08
Fix the build [plone 5]

ale-rt:


ale-rt:
This has to be cherry picked in to master

PR #138, opened 2018-05-09
The start page is also a form now

ale-rt:


ale-rt:
LOL this is part of #136

PR #142, opened 2018-05-15
Plone5 last modifier 1

ale-rt:


ale-rt:
PR should be mergeable, but we should investigate if we should use a separate table that logs the survey sessions edits.
This will give us history and insights on what happens.

PR #146, opened 2018-05-17
Move the security logic to the webhelpers view

ale-rt:
This will allow the security to be customized in other projects

### DIFF
--- a/src/euphorie/client/profile.py
+++ b/src/euphorie/client/profile.py
@@ -23,6 +23,7 @@ from euphorie.content.risk import IFrenchRisk
 from euphorie.content.risk import IRisk
 from euphorie.content.survey import ISurvey
 from five import grok
+from plone import api
 from sqlalchemy import sql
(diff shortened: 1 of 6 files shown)


src/euphorie/client/profile.py
```diff
@@ -341,33 +370,50 @@ def ProfileQuestions(self):
...
+            _(
+                u"message_field_required",
+                default=u"Please fill out this field."
+            ),
+            target_language=lang
+        )
         survey = aq_inner(self.context)
         self.profile_questions = self.ProfileQuestions()
-        self.session = SessionManager.session
         self.current_profile = extractProfile(survey, SessionManager.session)
```
pysailor: I'm not sure any more, but why is `SessionManager.session` used here instead of `self.session`?

ale-rt: I think it is something like a default... But now session is property, that means that it is evaluated whenever you need it and you will be more confident of its origin.

PR #156, opened 2018-08-01
 The account password is stored encrypted

ale-rt:
This PR adds:

1. event listeners to encrypt the passwords for a user account
2. an upgrade step to encrypt the passwords that are in plain text
3. verifies in various places in the code that the submitted password matched the encrypted one
4. replaces the password reminder functionality with a password reset one

### DIFF
--- a/src/euphorie/deployment/commands/upgradedb.py
+++ b/src/euphorie/deployment/commands/upgradedb.py
@@ -1,10 +1,12 @@
 # -*- coding: UTF-8 -*-
 ''' Upgrade the database tables if needed'''
+from datetime import datetime
 from euphorie.client import model
 from logging import getLogger
 from pkg_resources import get_distribution
 from pkg_resources import parse_version
 from Products.Five import zcml
+from sqlalchemy import sql
 from sqlalchemy.engine.reflection import Inspector
 from sys import argv
 from transaction import commit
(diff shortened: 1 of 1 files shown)


src/euphorie/deployment/commands/upgradedb.py
```diff
@@ -144,6 +154,7 @@ def main():
         add_group_id_to_session()
         add_published_to_session()
         add_last_modifier_id_to_session()
+        hash_passwords()
```
pysailor: @ale-rt Wow, looks like a great piece of work!
Just a suggestion: What if, instead of adding this as a general step to upgradedb, which could be run many times, you make an explicit upgrade step for this?
Then we would not need to worry about this comment, because the PW would only need to be encrypted once: either during the migration, or when the user sets / changes it. Does that make sense?

>         if BCRYPTED_PATTERN.match(password):
>             # The password is already encrypted, do not encrypt it again
>             # XXX this is broken with passwords that are actually an hash

ale-rt: Nope, that has to stay because the events are fired when a row in the account table is updated, regardless of the column.
If you remove that, changing any column will encrypt the password again.

For me the `hash_passwords` method should live in the upgradedb script, but if you have a better place I can move it.

pysailor: ok, thanks for the explanation.
So, in theory, after `hash_passwords` has run once on the thousands of accounts, it will be quick the next time that updatedb is run again, since the pattern-matching will quickly detect that there is nothing to do?

ale-rt: Yes

PR #160, opened 2018-09-05
Get the published context via get instead of getattr

ale-rt:


ale-rt:
@pysailor fixed

pysailor:
Great, thanks!

PR #161, opened 2018-09-06
When cloning a session copy some columns

ale-rt:


pysailor:
It works :) Both in plain Euphorie and the client-specific project. Thanks!

PR #181, opened 2019-08-22
Date picker i18n

ale-rt:
Added a ploneintranet inspired @@date-picker-i18n.json view that can be used
to provide translations for the calendar picker.

Example usage:
```
<input class="pat-date-picker"
       data-pat-date-picker="...; i18n: ${portal_url}/@@date-picker-i18n.json; ..."
/>
```

### DIFF
--- a/src/euphorie/client/browser/webhelpers.py
+++ b/src/euphorie/client/browser/webhelpers.py
@@ -20,6 +20,8 @@ from euphorie.content.survey import ISurvey
 from euphorie.content.utils import StripMarkup
 from euphorie.decorators import reify
 from euphorie.ghost import PathGhost
+from functools import partial
+from json import dumps
 from logging import getLogger
 from os import path
 from plone import api
@@ -38,10 +40,13 @@ from zope.component import getMultiAdapter
 from zope.component import getUtility
 from zope.component.hooks import getSite
 from zope.i18n import translate
+from zope.i18nmessageid import MessageFactory
 
 import Globals
 
 
+pl_message = MessageFactory("plonelocales")
+pae_message = MessageFactory("plone.app.event")
 logger = getLogger(__name__)
 
 
@@ -641,15 +646,19 @@ class WebHelpers(BrowserView):
             lang = "{0}_{1}".format(elems[0], elems[1].upper())
         return lang
 
+    @property
+    @memoize_contextless
+    def translate(self):
+        """ Simplify translations
+        """
+        target_language = self.translang()
+        return partial(translate, target_language=target_language)
+
     def closetext(self):
-        return translate(
-            _(u"button_close", default=u"Close"),
-            target_language=self.translang())
+        return self.translate(_(u"button_close", default=u"Close"))
 
     def email_sharing_text(self):
-        return translate(
-            _(u"I wish to share the following with you"),
-            target_language=self.translang())
+        return self.translate(_(u"I wish to share the following with you"))
 
     def getSecret(self):
         return getSecret()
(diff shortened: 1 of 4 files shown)


src/euphorie/client/browser/webhelpers.py
```diff
@@ -731,6 +740,66 @@ def show_logo(self):
...
+                        "weekday_wed_abbr",
+                        "weekday_thu_abbr",
+                        "weekday_fri_abbr",
+                        "weekday_sat_abbr",
+                    ]
+                ],
+            }
+        )
+        return json
+
```
pysailor: This seems to be very similar to the `date-picker-i18n.json` grok view that we already have:
https://github.com/euphorie/Euphorie/blob/master/src/euphorie/client/utils.py#L193

I am in favour of retiring grok :-) But could you please enhance this PR by removing the old grok view and adjusting all places where is `date-picker-i18n.json` is used?

ale-rt: I completely missed it

pysailor:
This will break for languages like nl-be. I will provide an update to fix this.

ale-rt:
Ok, the point behind that 

> This will break for languages like nl-be. I will provide an update to fix this.

I was trying to reuse this method: https://github.com/euphorie/Euphorie/blob/c867fce6ecc4886fe7fab8c3e38f74c6816ff4b0/src/euphorie/client/browser/webhelpers.py#L642-L647
Should that fixed as well?

pysailor:
I understood what you did, and for translations in the i18n:domain euphorie, that is fine, since here we do have a nl_BE .po file. But for the plonelocales i18n:domain, nl_BE does not exist, so the Belgians would see English month names if we did not use the simple form.

ale-rt:
So you are telling me that in some cases it is good to have the language variant and in some other it is not?

For example this thing is in the plone domain defined in Nuplone.
```
>>>
api.portal.translate("Solution", lang="nl_BE")
u'Oplossing'
```
and in this case it appears to be fine to have the language variant.

I agree that this is not the case with the language variant nl_BE in the domain plonelocales:
```
ipdb> api.portal.translate("month_jan", domain="plonelocales", lang="nl_BE")
u'January'
```
but it is not true in general.
What if one day Brazil wants to use this package? Will pt_BR support be broken?
```
ipdb> api.portal.translate("month_jan", domain="plonelocales", lang="pt_BR")
u'Janeiro'
```

So the real problem is that nl_BE is missing translations, I think we should just provide them and stick to the default translation machinery.
Maybe we can just copy those translations from the nl po file:
https://github.com/collective/plone.app.locales/blob/master/plone/app/locales/locales/nl/LC_MESSAGES/plonelocales.po

Also I would just start using directly `api.portal.get_current_language()` instead of custom methods.

pysailor:
Alessandro Pisa wrote on 23.08.19 09:41:
>
> So you are telling me that in some cases it is good to have the
> language variant and in some other it is not?
>
> For example this thing is in the plone domain defined in Nuplone.
>
> |>>> api.portal.translate("Solution", lang="nl_BE") u'Oplossing' |
>
> and in this case it appears to be fine to have the language variant.
>

Yes, because we have NuPlone in our hands and have therefore received
validated translation for Flemish (Belgian Dutch)


> I agree that this is not the case with the language variant nl_BE in
> the domain plonelocales:
>
> |ipdb> api.portal.translate("month_jan", domain="plonelocales",
> lang="nl_BE") u'January' |
>
> but it is not true in general.
> What if one day Brazil wants to use this package? Will pt_BR support
> be broken?
>
> |ipdb> api.portal.translate("month_jan", domain="plonelocales",
> lang="pt_BR") u'Janeiro'|


At the moment, indeed Brazlian would be broken, but only for the
calendar translations.


||
> ||
>
> So the real problem is that nl_BE is missing translations, I think we
> should just provide them and stick to the default translation machinery.
> Maybe we can just copy those translations from the nl po file:
> https://github.com/collective/plone.app.locales/blob/master/plone/app/locales/locales/nl/LC_MESSAGES/plonelocales.po
>

That is plain wrong in my opinion. By providing a nl_BE file, you would
suggest that you have a validated, country-specific translation for
Dutch. But we don't have that.
We only know that "regular" Dutch is sufficient for Flemish in this
case. We are not language experts for Flemish.

I am more surprised that the zope translation machinery does not fall
back on its own to the plain language variant, if the country-specific
version is not available. If you could find out how that might work, the
problem would by solved in my opinion.


> Also I would just start using directly
> |api.portal.get_current_language()| instead of custom methods.
>
Sounds good, as long as it works.

ale-rt:
Do you agree that the proper solution is to provide nl_BE translations for the plonelocales calendar related string?

pysailor:
Alessandro Pisa wrote on 23.08.19 10:07:
>
> Do you agree that the proper solution is to provide nl_BE translations
> for the plonelocales calendar related string?
>
No, as I said above.

ale-rt:
I thought your no was related to copying the translations from the nl po file.
You are anyway serving them with a trick in the code.

How can we proceed?
I would really stick to the default translations methods.

pysailor:
Alessandro Pisa wrote on 23.08.19 10:19:
>
> I thought your no was related to copying the translations from the nl
> po file.
> You are anyway serving them with a trick in the code.
>
> How can we proceed?
> I would really stick to the default translations methods.
>
I agree to use default methods like `api.portal.get_current_language()`.


I think the expected behavior should be:

- I request the translation for a country specific variant, like nl_BE
or pt_BR. If that translation is available, use it

- If that country specific translation is not available, fall back to
the basic language version (nl, pt, ...)

As far as I can see, that works already in general inside of Plone. We
have some tools with language "it-it". We don't support "it-it" in
euohorie, only "it" - but still the translations for "it" are used
correctly, since the translation machinery falls back to "it".

I think that we are just missing something very basic to get this right.

I do not think that the solution is to provide the country-specific
translations for every possible option.

PR #187, opened 2019-08-28
Session archival

ale-rt:
An archival date has been added and if it is in the past the session is
considered archived.
Edit permissions are lost and the session is removed from the dashboard
portlet.


ale-rt:
I added the hold merge because I want to move the check for the archived sessions in the `filter_valid_sessions` method.
This is a temporary fix, I think something more general and low level must the present to hide the archived sessions.
We already have two places where we are filtering for archived sessions (in the sidebar and in the dashboard).

src/euphorie/client/browser/webhelpers.py
```diff
@@ -664,13 +670,20 @@ def can_view_session(self):
...
+        )
 
     @property
     @memoize
     def can_publish_session(self):
         return self.can_edit_session
 
+    @property
+    @memoize
+    def can_archive_session(self):
```
ale-rt: @pysailor I bet here you want to also check for self.use_archive_feature

PR #188, opened 2019-08-28
Fetch image to be displayed on Start page in code (not in template)

pysailor:
Reason: make it easier to use different logic for fetching the image in customer package

### DIFF
--- a/src/euphorie/client/browser/session.py
+++ b/src/euphorie/client/browser/session.py
@@ -112,6 +112,15 @@ class SessionMixin(object):
         survey = aq_inner(self.context)
         return utils.HasText(getattr(survey, "introduction", None))
 
+    @property
+    @memoize
+    def scaled_tool_image_url(self):
+        if not getattr(self.survey, "external_site_logo", False):
+            return ""
+        scales = self.survey.restrictedTraverse("@@images")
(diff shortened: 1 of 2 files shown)


src/euphorie/client/browser/session.py
```diff
@@ -112,6 +112,19 @@ def has_introduction(self):
         survey = aq_inner(self.context)
         return utils.HasText(getattr(survey, "introduction", None))
 
+    @property
+    @memoize
+    def has_tool_image(self):
+        return bool(getattr(self.survey, "external_site_logo", None))
```
ale-rt: This method is redundant and should be removed IMHO, also because it can lead to wrong assumptions.
Infact reading the patch it seems that it is possible to have `self.has_tool_image == True` and `self.scaled_tool_image_url == ""` which seems to me something bad.

It is enough to check for the other one (`scaled_tool_image_url`) to return a truish valued.
Of course the line in that method:
```
if self.has_tool_image:
```
should be replaced with:
```
if not getattr(self.survey, "external_site_logo", False):
    return ""
...
```
This has also the benefit that the method will always return a string and that it will be less indented.

pysailor: Thanks for the quick review! I pushed the suggested changes.

PR #191, opened 2019-08-30
Remove unused code

ale-rt:
Remove the unused jsonify decorator that had a dependency on the
simplejson module that we do not want anymore


ale-rt:
I also suspect the dependency on the decorator package can be removed, But I would try a PR after this one has been merged

PR #197, opened 2019-09-02
Simplify code not using method is_outside_of_survey

ale-rt:


ale-rt:
I did not understand anything :)
Can you fix the PR or add a failing test?

src/euphorie/client/browser/webhelpers.py
```diff
@@ -387,6 +388,7 @@ def js_resources_url(self):
 
     @property
     @memoize
+    @deprecate("Replace with `not traversed_session`, deprecated in version 11.0.6.dev0")
     def is_outside_of_survey(self):
```
pysailor: I see now. This is semantically wrong. 

On the login form, we will always be outside of any traversed session. But, we might be in the context of a survey. Traversed session is something very different from survey. The traversed session cannot exist without a survey. But the survey can exist without the session.
I will instead try to refactor this method to make it more elegant. Will this be OK for you?

ale-rt: Oh yeah, I once again stumbled with the confusion between tool/survey/survey session//session.
Can this be good:
```
@deprecate("Replace with `not self.survey`, deprecated in version 11.0.6.dev0")
```

pysailor:
@ale I think I found a proper replacement: a check for `webhelpers/survey_url`. We only need it in login.pt. In the other templates that you identified, I think it's good to remove any use of `is_outside_of_survey`.
If the tests are still green, I think we can merge this now.

PR #203, opened 2019-09-06
Fix handling of Login and Register events inside a test session

pysailor:


src/euphorie/client/browser/webhelpers.py
```diff
@@ -653,6 +653,11 @@ def getSecret(self):
     def get_current_account(self):
         return get_current_account()
 
+    @memoize_contextless
+    def get_my_account_id(self):
+        account = self.get_current_account()
+        return account and account.id or ""
```
ale-rt: Just a suggestion: there is no need for introducing a new function given that in the pt you can do 
```
tal:define="account webhelpers/get_current_account" 
value="${account/id|nothing}"
```
It is a pity that we do not have an account property otherwise:
```
value="${webhelpers/account/id|nothing}"
```
was better.

pysailor: Thanks! Will use your suggestion

src/euphorie/client/browser/webhelpers.py
```diff
@@ -183,15 +183,15 @@ def redirectOnSurveyUpdate(self):
...
     @property
     @memoize
     def session_id(self):
-        return getattr(self.traversed_session.session, 'id', '')
+        traversed_session = self.traversed_session
+        return (
+            getattr(self.traversed_session.session, "id", "")
+            if traversed_session
+            else ""
+        )
```
ale-rt: This is over protection:
```
return getattr(self.traversed_session.session, "id", "")
```
seems enough to me.

pysailor: But what if `self.traversed_session` is None? This happens e.g. on the dashboard. If any template that is used outside of a traversed session calls `session_id`, we would get an error.
(I know that such templates should not be calling `session_id` on the first place. So maybe it IS better to fail.

ale-rt: I see now that this is different...

Anyway it should be good to:
```
return traversed_session.session.id if traversed_session else ""
```

src/euphorie/client/browser/login.py
```diff
@@ -79,15 +79,19 @@ def login(self, account, remember):
...
+        if not account_id:
             return
         account = getSecurityManager().getUser()
-        session = Session.query(model.SurveySession).get(session_id)
-        session.account_id = account.id
-        return session
+        sessions = (
+            Session.query(model.SurveySession)
+            .filter(model.SurveySession.account_id == account_id)
+            .all()
```
ale-rt: Just a suggestion: calling all here is not needed

PR #205, opened 2019-09-09
Risk image

ale-rt:
This is based on PR #204 to run the upgrade steps to create the missing columns.
You may decide to merge this one and ignore #204

There are some weird injection problems that cause unpredicted behavior, see the related ticket https://github.com/syslabcom/oira.prototype/issues/141


versions.cfg
```diff
@@ -10,7 +10,9 @@ z3c.recipe.scripts = 1.0.1
 plone.recipe.zope2instance  = 4.4.0
 
 # Euphorie dependencies
+alembic = 1.1.0
```
pysailor: Do we need to add 
`python-editor = 1.0.4`
as well? It is a dependency of alembic.

ale-rt: I missed it, thanks for noting it, I added this in d978ce2

src/euphorie/deployment/upgrade/alembic/versions/24_.py
```diff
@@ -0,0 +1,33 @@
...
+        "risk", sa.Column("image_data_scaled", sa.LargeBinary(), nullable=True)
+    )
+    op.add_column("risk", sa.Column("image_filename", sa.UnicodeText(), nullable=True))
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("risk", "image_data")
+    op.drop_column("risk", "image_filename")
```
pysailor: Just a doubt (not important): the downgrade should probable also drop column `image_data_scaled`?

ale-rt: I think we will never used downgrades but anyway I pushed the requested change

PR #209, opened 2019-09-19
Make it possible to pass in a scale name to the omega scaling

pysailor:
NOTE: we still only have one column in postgres for storing the scaled
iamge. That means, whichever scale is requested first, wins.
So, for the moment, we need to make sure to always request the same
correct scale when calling this.

### DIFF
--- a/src/euphorie/client/browser/risk.py
+++ b/src/euphorie/client/browser/risk.py
@@ -502,7 +502,12 @@ class ImageDisplay(DisplayFile):
     ../@@image-display/image_large/${here/image_filename}
     """
 
-    def get_or_create_image_scaled(self):
+    @property
+    @memoize
+    def allowed_sizes(self):
+        return getAllowedSizes()
+
+    def get_or_create_image_scaled(self, scale_name="large"):
         """ Get the image scaled
         """
         if self.context.image_data_scaled:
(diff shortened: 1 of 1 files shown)


src/euphorie/client/browser/risk.py
```diff
@@ -530,8 +535,9 @@ def _getFile(self):
         if self.context.image_data is None:
             raise NotFound(self, self.fieldname, self.request)
 
-        if self.fieldname == "image_large":
-            image_data = self.get_or_create_image_scaled()
+        requested_scale = self.fieldname.partition("_")[-1] if self.fieldname else None
+        if requested_scale in self.allowed_sizes:
+            image_data = self.get_or_create_image_scaled(requested_scale)
```
ale-rt: I do not think this is right because it has a race condition.
If you first request the "large" scale you will store it in `self.context.image_data_scaled`, then if you request the "preview" scale you will get the "large" one already stored in `self.context.image_data_scaled`.
Viceversa if you first request the "preview" and then the "large" you will get the "preview" one.
So I think we also need a different column to store a different preview.

With this in mind `image_data_scaled` seems a bad name...

pysailor: ok, then I will change it to just always use the "training" scale

pysailor: Your comment addresses exactly the problem that I wrote in my comment :)

ale-rt: > ok, then I will change it to just always use the "training" scale
I would not hardcode the name.
I am thinking more to:
1. check if "requested_scale" if in the allowed sizes
2. check if we have a column where to store it
3. if the scale is allowed and we have a column return the column data or create the scale, otherwise return the unscaled image (or raise NotFound, depending on what you prefer)

pysailor:
For sake of simplicity, I close this, and opened #210 instead

ale-rt:
Cool :) I like simplicity. 
Let's keep in mind the challenges we had now if we want a more refined image scaling story.

PR #212, opened 2019-09-20
WIP It fixes

pysailor:


src/euphorie/client/browser/risk.py
```diff
@@ -348,7 +352,12 @@ def _prepare_risk(self):
 
         # Italian special
         if self.webhelpers.country == "it":
-            self.skip_evaluation = True
+            if self.risk and (
+                self.risk.type == "top5" or self.risk.evaluation_method == "fixed"
+            ):
+                self.skip_evaluation = True
+            else:
+                self.evaluation_condition = "condition: answer=no or answer=yes"
```
ale-rt: Can `skip_evaluation` and `evaluation_condition` those be properties?
Setting class attributes on the fly is hard to debug and also this could shave some lines from the very very long `_prepare_risk` method.

src/euphorie/client/browser/risk.py
```diff
@@ -750,28 +763,33 @@ def __call__(self):
...
         # Italian special
+        if self.is_custom_risk:
+            self.risk.description = u""
+            self.risk.evaluation_method = u""
         if self.italy_special:
-            self.skip_evaluation = True
+            if self.risk and (
+                self.risk.type == "top5" or self.risk.evaluation_method == "fixed"
+            ):
+                self.skip_evaluation = True
```
ale-rt: Can `skip_evaluation` be a property?

src/euphorie/client/browser/risk.py
```diff
@@ -750,28 +763,33 @@ def __call__(self):
...
+                    getattr(solution, "prevention_plan", "") or "").strip()
+                match = description
                 if measures_full_text:
-                    match = u"%s: %s" % (
-                        (getattr(solution, "description", "") or "").strip(),
-                        (getattr(solution, "prevention_plan", "") or "").strip(),
-                    )
-                else:
-                    match = solution.description.strip()
+                    if prevention_plan:
```
ale-rt: Minor: join the ifs -> `if measures_full_text and prevention_plan:`.
Implies less indentation

src/euphorie/client/browser/webhelpers.py
```diff
@@ -279,15 +279,6 @@ def logoMode(self):
     @memoize
     def styles_override(self):
```
ale-rt: Can this be removed?

src/euphorie/client/docx/compiler.py
```diff
@@ -502,7 +501,7 @@ def compile(self, data):
         self.set_consultation_box()
 
 
-class DocxCompilerItaly(DocxCompiler):
+class DocxCompilerItalyOriginal(DocxCompiler):
```
ale-rt: Can this be removed? I do not see it referenced anywhere else in this PR.

src/euphorie/client/docx/compiler.py
```diff
@@ -621,27 +647,43 @@ def set_cell_actions(self, cell, risk):
...
                 )
                 paragraph.runs[0].italic = True
             if action.get('planning_start', None):
                 paragraph = cell.add_paragraph(
-                    u"Date de fin: {}".format(action['planning_start']),
+                    u"{0} {1}".format(
+                        api.portal.translate(
+                            _(u"report_end_date", default=u"To be done by:")
+                        ),
+                        action['planning_start']),
```
ale-rt: It would be better to add an i18n name in the translated string:
```
api.portal.translate(
    _(u"report_end_date", default=u"To be done by: ${date}", mapping={"date": action['planning_start'])})
)
```
Untested ^

src/euphorie/client/docx/compiler.py
```diff
@@ -670,56 +712,61 @@ def set_modules_rows(self, data):
             if not risks:
                 continue
             row_module = table.add_row()
+            cell = row_module.cells[0]
+            cell.paragraphs[0].text = module.get('title', '')
+            cell.vertical_alignment = WD_ALIGN_VERTICAL.CENTER
+            self.set_row_borders(row_module)
+            count = 0
             for r_idx, risk in enumerate(risks):
```
ale-rt: Why not just for `risk in risks`?

src/euphorie/client/docx/compiler.py
```diff
@@ -732,13 +779,82 @@ def compile(self, data):
         Check the file .../daimler/oira/tests/mocked_data.py
         to understand its format
         '''
+        self.unanwered_risks = []
+        self.not_applicable_risks = []
```
ale-rt: Can `unanwered_risks` and `not_applicable_risks` be properties?

src/euphorie/client/browser/risk.py
pysailor: ok, done

src/euphorie/client/browser/risk.py
pysailor: ok, done

src/euphorie/client/browser/risk.py
pysailor: ok, done

src/euphorie/client/browser/webhelpers.py
pysailor: ok, done

src/euphorie/client/docx/compiler.py
pysailor: Thanks for the hint, I fixed it, now it's used again.

src/euphorie/client/docx/compiler.py
```diff
@@ -621,27 +647,43 @@ def set_cell_actions(self, cell, risk):
...
                 run.text = _simple_breaks(action['requirements']),
 
             if action.get('responsible', None):
                 paragraph = cell.add_paragraph(
-                    u"Responsable: {}".format(action['responsible']),
+                    u"{0} {1}".format(
+                        api.portal.translate(
+                            _(u"report_responsible", default=u"Responsible:")
+                        ),
+                        action['responsible']),
```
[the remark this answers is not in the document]
pysailor: ok, done

src/euphorie/client/docx/compiler.py
pysailor: ok, done

src/euphorie/client/docx/compiler.py
pysailor: ok, done

src/euphorie/client/docx/compiler.py
```diff
@@ -670,56 +712,61 @@ def set_modules_rows(self, data):
...
 
-        # # Set the footer row of the table
-        # row = table.add_row()
-        # self.set_row_borders(row)
-
         def _merge_cells(row):
-            cell1, cell2, cell3, cell4 = row.cells
-            cell2_new = cell2.merge(cell3)
-            cell2_new = cell2_new.merge(cell4)
+            """
```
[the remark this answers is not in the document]
pysailor: ok, done

src/euphorie/content/risk.py
```diff
@@ -548,12 +548,12 @@ def get_pre_defined_measures(self, request=None):
             description = item.description and item.description.strip() or ""
             prevention_plan = (
```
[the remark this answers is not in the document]
pysailor: will probably be refactored anyways, I skipped it for the moment

src/euphorie/client/docx/compiler.py
pysailor: What would be the benefit? Those lists get filled in `set_modules_rows`. Do you think a property with a setter would be better?

pysailor:
Thanks for the review! I fixed most of the issues that you mentioned.

src/euphorie/client/docx/compiler.py
ale-rt: The benefit is code clarity.
Why those attributes are defined here if they are filled in `set_modules_rows`?
I might be missing something trivial but code like this smells to me.

src/euphorie/client/docx/compiler.py
```diff
@@ -732,13 +787,82 @@ def compile(self, data):
         Check the file .../daimler/oira/tests/mocked_data.py
         to understand its format
         '''
+        self.unanwered_risks = []
```
ale-rt: This should be `unanswered_risks`.

src/euphorie/client/docx/compiler.py
ale-rt: > So do you think that writing a property method with a setter would be appropriate here? 

No

> Or do you have another suggestion?

I do not know, it seems to me that the `set_modules_rows` is overly complex.
I am wondering why, for example, the data structure you are passing does not already have this partition.
But anyway, I do not want to block your work.

pysailor: The data structure is created here: https://github.com/euphorie/Euphorie/blob/fb16d0be4eb3b9c5d6d41f30fc1ef685793d1944/src/euphorie/client/docx/views.py#L84

That means, for every module, it lists all risks in their given order. Of course, I could already create the partition in the method `get_risks_for`, so that the return value might be a tuple (risks, unanswered_risks, not_applicable_risks). 
But then, `set_modules_rows` would become even more complex. If `only_anwered_risks` is False, then I would need to iterate both over risks and unanswered_risks, and make sure that I put them in the correct order in the report. (e.g. 1.1. is in `risk`, 1.2 is in `unanswered_risks`, 1.3 is in `risks` again...)

For me, `set_modules_rows` is a logical place to make this distinction: iterate over all the risks in the module, and decide 
1) not applicable? -> skip in the normal report, and save for later 
2) not answered? -> skip in the normal report if we only want risks with answers, and save for later.

I'm trying to be as modular as possible, to support various "flavors" of the report. E.g. the French want to show not answered risks in the normal report, the Italians want to show it in a separate list.

I'm sure there is a better way of coding this, but for the moment, I think it can only be achieved with a substantial re-write of the whole docx module. 
In terms of long-term maintenance, it will probably be worth the effort, especially if you find the current state confusing. I would first start by spending a day or two in writing tests.

pysailor: I have an idea how to make it clearer: I will not define `unanwered_risks` and `not_applicable_risks` as class variables, but as return values of `set_modules_rows`, so that they can then be passed to `add_extra` in `compile`.

pysailor:
I did something stupid in e8418aee7804b0142f8f9f84b486a3bacccc9623
Will fix that tomorrow...

src/euphorie/client/docx/views.py
```diff
@@ -158,6 +157,7 @@ def get_risks_for(self, sql_module):
                 u'measures': measures,
                 u'epilogue': u'',
                 u'justifiable': sql_risk.identification,
+                u'number': sql_risk.number,
```
ale-rt: Should we just add to the dict the `sql_risk` and then evaluate the number when needed?

pysailor:
Just discussed with @ale-rt: we want to find a better solution than the various `utils.setLanguage(self.request, self.survey, self.survey.language)` call, possibly by re-activating the survery traverser. But that will be in a new PR...

PR #213, opened 2019-09-23
Publication change

pysailor:
1) Don't record publication as a modification event
2) Show _who_ published a session

### DIFF
--- a/src/euphorie/client/browser/templates/publication_menu.pt
+++ b/src/euphorie/client/browser/templates/publication_menu.pt
@@ -18,7 +18,8 @@
       <div class="pat-message notice"
          tal:condition="last_modifier"
       >
-        <p tal:condition="session/published">
+        <p tal:condition="session/published"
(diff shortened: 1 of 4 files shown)


src/euphorie/client/browser/templates/publication_menu.pt
```diff
@@ -26,6 +27,8 @@
                   tal:define="value session/published/isoformat|nothing;"
                   datetime="${value}"
             >${value}</time>
+            <tal:i18n i18n:translate="by">by</tal:i18n>
+            <a href="mailto:${last_publisher/email}">${last_publisher/email}</a>
```
ale-rt: `last_publisher` is defined as `session/last_publisher|nothing`.
Maybe wrapping the new html with a `tal:condition="last_publisher"` might be good, but I am not sure: if is published we should have a `last_publisher`.
Anyway I would move the definition inside the `<p tal:condition="session/published">` because otherwise we will define a variable that will not be used.
This is anyway a real minor issue.

PR #216, opened 2019-10-01
image upload omega: only process a file that PIL recognizes as image

pysailor:


src/euphorie/client/browser/risk.py
```diff
@@ -503,11 +503,20 @@ def calculatePriority(self, risk, reply):
 
 class ImageUpload(BrowserView):
 
+    def _return(self):
```
ale-rt: Maybe a name like `redirect` or `_redirect` would be better, but this is a really minor thing

src/euphorie/client/browser/risk.py
```diff
@@ -503,11 +503,20 @@ def calculatePriority(self, risk, reply):
...
+            "{}/@@identification".format(self.context.absolute_url())
+        )
+
     def __call__(self):
         if self.request.form.get("image"):
             image = self.request.form["image"]
             new_data = image.read()
             if self.context.image_data != new_data:
+                try:
+                    PIL.Image.open(BytesIO(new_data))
```
ale-rt: Don't we want to inform the user about what just happened?

src/euphorie/client/browser/risk.py
ale-rt: Just in case you like consistency:
https://github.com/euphorie/Euphorie/blob/d81c67d679d0eaf5dd08583579a172f1bc99a5db/src/euphorie/client/browser/reset_password.py#L51-L56

pysailor: Thanks! I knew we already had something like that, and I also did not like the name `_return` either... I this case, I like consistency a lot, will change! :-)

src/euphorie/client/browser/risk.py
pysailor: Yeah, I thought about that too. As a MVP I think it might be good enough to simply not display anything. If we emit a message like "Not a valid image", we need translation again for 20 languages.
But probably, it is better to display a warning. I'll add something.

PR #218, opened 2019-10-04
Most countries only have one language, use that as a default

pysailor:


src/euphorie/client/country.py
```diff
@@ -43,6 +42,28 @@ class ClientCountry(dexterity.Container):
...
+        "lv": "lv",
+        "lt": "lt",
+        "pt": "pt",
+        "si": "sl",
+    }
+
+    @property
+    def language(self):
+        if self.id in self.language_mapping_by_country:
+            return self.language_mapping_by_country[self.id]
```
ale-rt: Isn't this equivalent to:
```
return self.language_mapping_by_country.get(self.id)
```
?

pysailor: No, far from it!
Examples: Country Cyprus (cy) has language Greek (el), Also country Greece (gr) has language Greek (el). Country Slovenia == si, has language Slovenian == sl.
Or, if we had Austria (at), it would be German (de).
There are many more examples where the country id is not the same as the language code.

pysailor: Oh, I misread your question :)
Yes, you are right.

PR #224, opened 2019-10-16
Refine sessions query

pysailor:
When fetching sessions: Make it possible to filter on group independently whether we filter on the current account or not


pysailor:
Obsolete, superseded by #225

PR #225, opened 2019-10-17
feat: better account and group filters

ale-rt:
BREAKING CHANGE: the include_group parameter is removed and replaced
with filter_by_group parameter which has more flexibility

This PR makes it possible to optionally filter sessions by account or group,
it is also possible to filter for lists of users or groups.

Refs. and replaces #224


ale-rt:
Note some improvements in the docstring for `get_account_filter` and `get_group_filter` should be welcome but I am short in time right now.

PR #232, opened 2019-11-07
feat: messages on the start page are customizable

ale-rt:


ale-rt:
Hold merge... there is complication...

src/euphorie/client/browser/session.py
```diff
@@ -164,6 +164,38 @@ def message_required(self):
...
+            message = _(
+                """
+                    You don't have edit rights for this risk assesment,
+                    but you can download this risk assessment in various forms in the
+                    <a href="${absolute_url}/@@report_view"
+                        i18n:name="download_section"
+                    >download section</a>.
+                """,
+                mapping={"absolute_url": self.context.absolute_url()},
+            )
```
pysailor: This results in a broken .pot file. I will try to fix this

pysailor: Ok, I think I fixed it. Not sure if it could also be done in a simpler way ;) But this pattern has worked before, at least.

ale-rt: I think it is broken as well as my original code :)
the fact is that `i18n:name="download_section"` should not even be there.

pysailor: Yes, that's why I removed the `i18n:name="download_section"`. And I don't want to have any markup in my po file, so I need to simulate the effect that a `i18n:name` has in a page template.

This keeps the 2 original message strings:

* msgid "You don't have edit rights for this risk assesment, but you can download this risk assessment in various forms in the ${download_section}."
* msgid "download section"

But of course we gain the additional garbage

PR #233, opened 2019-11-12
In the `copySessionData` method of SurveySession, add a hook at the end

pysailor:
that allows to run further updates


pysailor:
Actually, screw this, I will use `super` instead.

PR #234, opened 2019-11-15
feat: risk template customizable

pysailor:
Move the computation of the template to be used into a separate method, so that it can be customised.

### DIFF
--- a/src/euphorie/client/browser/risk.py
+++ b/src/euphorie/client/browser/risk.py
@@ -52,7 +52,8 @@ class IdentificationView(BrowserView):
     """A view for displaying a question in the identification phase
     """
 
-    template = ViewPageTemplateFile("templates/risk_identification.pt")
+    default_template = ViewPageTemplateFile("templates/risk_identification.pt")
+    custom_risk_template = ViewPageTemplateFile("templates/risk_identification_custom.pt")
     variation_class = "variation-risk-assessment"
 
     question_filter = None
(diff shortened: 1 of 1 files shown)


src/euphorie/client/browser/risk.py
```diff
@@ -52,7 +52,7 @@ class IdentificationView(BrowserView):
     """A view for displaying a question in the identification phase
     """
 
-    template = ViewPageTemplateFile("templates/risk_identification.pt")
+    default_template = ViewPageTemplateFile("templates/risk_identification.pt")
     variation_class = "variation-risk-assessment"
```
ale-rt: I would add here a:
```
custom_risk_template = ViewPageTemplateFile("templates/risk_identification_custom.pt")
```

src/euphorie/client/browser/risk.py
```diff
@@ -261,18 +261,23 @@ def __call__(self):
         else:
             self._prepare_risk()
             if self.is_custom_risk:
```
ale-rt: If my previous suggestion is accepted I would write:
```
if not self.is_custom_risk:
    return self.default_template()
```
and dedent the code one level

src/euphorie/client/browser/risk.py
```diff
@@ -261,18 +261,23 @@ def __call__(self):
...
-                    self, ""
-                )  # noqa
                 next = FindNextQuestion(
                     self.context, self.context.session, filter=self.question_filter
                 )
                 self.has_next_risk = next or False
-            else:
-                template = self.template
-            return template()
+            return self.template()
```
ale-rt: If my previous suggestions are accepted this can be:
```
return self.custom_risk_template.__get__(self, "")() 
```
Actually I think this might also work:
```
return self.custom_risk_template() 
```
Not sure what `.__get__(self, "")` does...

src/euphorie/client/browser/risk.py
```diff
@@ -261,18 +261,23 @@ def __call__(self):
...
+
+    @property
+    def template(self):
```
ale-rt: Given my previous suggestions are accepted this property can go away

pysailor: But, I _want_ to have this property, so that I can overwrite it in daimler, so that I can consider more use cases and add a 3rd template option.

Your other suggestions are fine. I will try them out & add them.

PR #235, opened 2019-11-23
Rearrange training completely, using browser view on the context

pysailor:


src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+
+    @property
+    @memoize
+    def existing_measures(self):
+        try:
+            existing_measures = loads(self.context.existing_measures)
+            # Backwards compat. We used to save dicts in JSON before we
+            # switched to list of tuples.
+            if isinstance(existing_measures, dict):
+                existing_measures = [(k, v) for (k, v) in existing_measures.items()]
```
ale-rt:
```
existing_measures = existing_measures.items()
```
is the same as:
```
existing_measures = [(k, v) for (k, v) in existing_measures.items()]

File: src/euphorie/client/browser/training.py [hunk 2]
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+            existing_measures = [
+                text.strip() for (text, active) in existing_measures if active
+            ]
+        except:
+            existing_measures = []
```
ale-rt: I would not catch a bare exception without at least logging something...
can we add a:
```
logger.exception("Something is broken in %r", self.context)
```

src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+        return existing_measures
+
+    @property
+    @memoize
+    def image(self):
+        if self.is_custom:
+            if getattr(self.context, "image_data", None):
```
ale-rt: Anti if:
```
if getattr(self.context, "image_data", None):
    return
```

src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+                _view = self.context.__of__(
```
ale-rt: It seems strange to me that this is needed

src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+                    self.webhelpers.traversed_session.aq_parent["custom-risks"]
+                ).restrictedTraverse("image-display")
+                _view.field_name = "image_large"
+                return _view.get_or_create_image_scaled()
+            return None
+        image = self.zodb_elem.image and self.zodb_elem.image.data or None
```
ale-rt: It seems to me there is no point in getting the image if `self.for_download` is False.

src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+        if image and self.for_download:
+            try:
+                scales = self.zodb_elem.restrictedTraverse("images", None)
+                if self.item_type == "module":
```
ale-rt: What is the point of computing the scale_name if scales is falsish?

src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+                    scale_name = "training_title"
+                else:
+                    scale_name = "training"
+                if scales:
+                    scale = scales.scale(fieldname="image", scale=scale_name)
+                    if scale and scale.data:
+                        image = scale.data.data
+            except:
+                image = None
```
ale-rt: I would not catch a bare exception without at least logging something...

src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+        return "for_download" in self.request
+
+    @property
+    @memoize
+    def title_image(self):
+        survey = self.context.aq_parent
+        return (
+            getattr(survey, "external_site_logo", None)
+            and survey.external_site_logo.data
+            or None
```
ale-rt: It seems a case were a try/except should be preferred to a getattr
```
try:
     return self.context.aq_parent.external_site_logo.data
except AttributeError:
     return
```

src/euphorie/client/browser/training.py
```diff
@@ -12,43 +19,207 @@ class TrainingSlide(BrowserView):
...
+        )
+
+    @property
+    @memoize
+    def slide_data(self):
+        modules = self.getModulePaths()
+        risks = self.getRisks(modules, skip_unanswered=self.skip_unanswered)
```
ale-rt: Looks strange to pass an instance attribute (`self.skip_unanswered`) to an instance method.

src/euphorie/client/browser/training.py
pysailor: I think you mean `if not getattr(self.context, "image_data", None)` ?

ale-rt: yes, not ;)

pysailor:
Thanks for the review, @ale-rt ! I made the suggested simplifications. I'll merge once the tests are green again.

PR #238, opened 2019-12-04
Improve stats data

pysailor:
* Save creation date of accounts
* Default account type is "full" instead of None


pilz:
Wonderful!!!

src/euphorie/deployment/upgrade/alembic/versions/25_.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = '24'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('account', sa.Column('created', sa.DateTime(), nullable=True))
+    op.execute("UPDATE account set account_type='full' WHERE account_type is Null")
+    op.add_column('group', sa.Column('brand', sa.String(length=64), nullable=True))
```
ale-rt: Uhm, this line smells:
```
op.add_column('group', sa.Column('brand', sa.String(length=64), nullable=True))
```
I think you should remove it

src/euphorie/client/browser/login.py
```diff
@@ -285,6 +285,7 @@ def _tryRegistration(self):
             account.loginname = loginname
             account.password = reply.get("password1")
             account.account_type = config.CONVERTED_ACCOUNT
+            account.created = datetime.datetime.now()
```
ale-rt: What will be the created value for existing users?

src/euphorie/deployment/upgrade/alembic/versions/25_.py
pysailor: Hm, right. we don't need brand in Euphorie.

src/euphorie/client/browser/login.py
pysailor: Null

src/euphorie/deployment/upgrade/alembic/versions/25_.py
ale-rt: More context: it should be already in the DB https://github.com/euphorie/Euphorie/blob/331630d13d0a2577e3f6e0351540248ba75650ba/src/euphorie/deployment/commands/upgradedb.py#L90

src/euphorie/client/browser/login.py
pysailor: Unless @pilz has a differen opinion?

src/euphorie/client/model.py
```diff
@@ -398,15 +398,21 @@ class Account(BaseObject):
     password = schema.Column(types.Unicode(64))
     tc_approved = schema.Column(types.Integer())
     account_type = schema.Column(
-        Enum([u"guest", u"converted", None]),
-        default=None,
+        Enum([u"guest", u"converted", "full"]),
```
ale-rt: This is missing an upgrade step, right?

pysailor: It should already be present, in the alembic upgrade: https://github.com/euphorie/Euphorie/pull/238/files#diff-fbef39df29989bc254ed99389d7fad78R21

pysailor:
@ale-rt What do you think is still missing in this PR?

src/euphorie/client/model.py
ale-rt: When you change an Enum (AFAIK) you have to update the type.
See https://stackoverflow.com/questions/1771543/adding-a-new-value-to-an-existing-enum-type

ale-rt:
As mentioned I think an update for the Enum type is necessary (I have bad memories from the pat)

pysailor:
Ok, thanks. I will do some research on this Enum. At the moment, I cannot imagine what else I can update besides what alembic already does.

pysailor:
@pilz as you can see, no quick win here

ale-rt:
I you are in a hurry merge this

pysailor:
No I am not in a hurry. I thought I had understood the problem at hand by providing a migration for the "account_type" values. But apparently there is more to this issue.

I don't really have a strong interest in this change. @pilz wants the regular members to have the account type "full". He is probably not aware of the consequences either.

ale-rt:
Just to clarify:
- I remember that years ago I had a problem with enums on postgres (after adding a value I had to perform a DB migration that sounds like the one described here https://stackoverflow.com/questions/1771543/adding-a-new-value-to-an-existing-enum-type)
- the note above the most accepted answer says some more convenient way of doing that is present since Postgres 9.1
- I also found: https://stackoverflow.com/questions/14845203/altering-an-enum-field-using-alembic

Said that if everything works for you I am not blocking this.

pysailor:
Hi @ale-rt 
I did some more checking and educated myself a bit about `enum`.
In postgres, it is possible to create an Enum Type: https://www.postgresql.org/docs/9.1/datatype-enum.html
If I follow the simple example that is given, I get the following table:

    euphorie=# \d person
                    Table "public.person"
        Column    | Type | Collation | Nullable | Default 
    --------------+------+-----------+----------+---------
     name         | text |           |          | 
     current_mood | mood |           |          | 

So the type of the `current_mood` column is my ENUM type. And if I try to insert a value that is not supported, I get an error:

    euphorie=# INSERT INTO person VALUES ('Larry', 'exctatic');
    ERROR:  invalid input value for enum mood: "exctatic"

On the other hand, our Account table has the following structure:

	euphorie=> \d account
	                                    Table "public.account"
	    Column    |          Type          |                      Modifiers
	--------------+------------------------+------------------------------------------------------
	 id           | integer                | not null default nextval('account_id_seq'::regclass)
	 loginname    | character varying(255) | not null
	 password     | character varying(64)  |
	 tc_approved  | integer                |
	 account_type | character varying(16)  |
	 group_id     | character varying(32)  |

The `account_type` column is of type varchar. No value checking is performed in postgres.

	euphorie=# insert into account VALUES (99999, 'nobody', '', 0, 'bogus_type');
	INSERT 0 1
	euphorie=# select * from account order by ID desc limit 1;
	  id   | loginname | password | tc_approved | account_type | group_id | created 
	-------+-----------+----------+-------------+--------------+----------+---------
	 99999 | nobody    |          |           0 | bogus_type   |          | 

Instead, the value checking happens in python:
https://github.com/euphorie/Euphorie/blob/master/src/euphorie/client/enum.py#L88
And the mentioned `self.values` is set in the `__init__` method when the instance starts, based on what gets passed via the constructor https://github.com/euphorie/Euphorie/pull/238/files#diff-183ec4f73c3b9310273c20ff3b8da310R401
And indeed, after my change, I saw test failures when I tried to set an account_type to None, since it was no longer supported by the Enum.

So, I think nothing needs to happen in postgres that is not already taken care of.

ale-rt:
LOL, now I see... 
That is a custom Enum type and not the sqlalchemy one https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Enum (which provides python and DB validation).
So evaulate yourself what you want to do here. (I am not a fan of Enum, anyways)

PR #240, opened 2019-12-09
More Info about a tool

pysailor:
According to proto, the additional Info about a tool, incl a potentially present image, is not shown directly on the Preparation page any more, but instead in a modal


src/euphorie/client/browser/session.py
```diff
@@ -222,6 +222,11 @@ def update(self):
             )
 
 
+class MoreInfo(SessionMixin):
```
ale-rt: Just use the `SessionMixin` class in the zcml. No need for an extra class right now

src/euphorie/client/browser/templates/start.pt
```diff
@@ -51,6 +51,13 @@
           <section tal:condition="context/description">
             <p>${context/description}</p>
           </section>
+          <p>
+              <a class="show-more pat-modal"
+                 href="${here/absolute_url}/@@tool-more-info#document-content"
+                 data-pat-modal="class: small"
+                 i18n:translate=""
+                 >Learn more about this tool…</a>
```
ale-rt: I would prefer `&hellip;` instead of `…`

src/euphorie/client/browser/templates/tool-more-info.pt
```diff
@@ -0,0 +1,92 @@
...
+      metal:use-macro="context/@@modal-template/macros/shell"
+      i18n:domain="euphorie"
+>
+
+  <body>
+    <metal:slot fill-slot="content">
+      <div class="pat-modal"
+           id="document-content"
+      >
+        <h1 class="panel-class-">
```
ale-rt: Is this needed?

src/euphorie/client/browser/templates/tool-more-info.pt
```diff
@@ -0,0 +1,92 @@
...
+
+
+                <section class="pat-rich tool-info" id="about-tool">
+<tal:comment condition="nothing">
+Will be activated when we re-build the Help section
+                  <h2 i18n:translate="">About this tool</h2>
+</tal:comment>
+
+                <tal:description condition="context/description">
+                  <p>${context/description}</p>
```
ale-rt: Too verbose, I would just add the tal:condition inside the p element

src/euphorie/client/browser/templates/tool-more-info.pt
pysailor: I guess so, it's in proto. I will leave it in.

PR #243, opened 2020-02-03
Progress Indicator

reinhardt:
Completion percentage (answered risks / total risks) is recorded on the session and used to display a progress indicator.

See https://github.com/syslabcom/oira.prototype/blob/88b9f58090e49ed9f9b3e002940d671ba25421e9/_layouts/status.html#L95

`update_completion_percentage()` does a lot of calls to `query()` which will affect performance. The separate calls can probably be transformed into one or two complicated ones. Is that worth the effort? Or is there another way to optimize?


src/euphorie/client/tests/test_survey.py
```diff
@@ -167,3 +167,107 @@ def testSetupContext_TwoSteps(self):
...
+        )
+        self.q2 = self.mod21.addChild(
+            model.Risk(title=u'question 2', risk_id='2', zodb_path='k/c/u')
+        )
+        self.q1.postponed = False
+        self.q1.identification = "yes"
+        self.mod21.skip_children = True
+
+        self.survey.update_completion_percentage()
+        self.assertEqual(self.survey.completion_percentage, 100)
```
pysailor: Please add a test that proves that if the module mod2 is optional and false, that the risks in submodule m21 are not counted.

src/euphorie/client/model.py
```diff
@@ -1031,6 +1038,38 @@ def tool(self):
...
+                total_risks_query
+                .filter(Risk.identification != None)
+            )
+            answered_risks = answered_risks + answered_risks_query.count()
+
+        self.completion_percentage = (
+            int(round((answered_risks / total_risks) * 100.))
+            if total_risks else 0.
+        )
+
```
pysailor: I think that this approach will not be able to deal with all complexities in OiRA. Example: If a an optional module is turned off, and this module has submodules, will the risks inside the submodules be correctly discarded from the equation? (See below my comment on the test)

And it seems to me that we need an awful number of queries. Let me see if we might have examples that combine some of the steps into a single query.

src/euphorie/client/templates/shell.pt
```diff
@@ -271,11 +271,15 @@
                      data-pat-switch="selector: body; remove: focus-*; add: focus-document &amp;&amp; selector: body; remove: sidebar-on; add: sidebar-off"
                      tal:condition="not:disabled"
                      i18n:translate="navigation_status"
-                  >Status</a>
+                  >Status
+                    <strong class="pie-graph" data-percentage="${webhelpers/traversed_session/session/completion_percentage|string:0}" title="${webhelpers/traversed_session/session/completion_percentage|string:0}% Complete">${webhelpers/traversed_session/session/completion_percentage|string:0}%</strong>
```
pysailor: i18n needs to be handled, so that a German user will see "69% Komplett" instead of "69% Complete" etc

pysailor: Hehehe, I just realised, this whole tag is not valid:
Note that the outer `<a>` tag already has an `i18n:translate` statement. In tools of a different language, the percentage indicator simply gets overwritten by the translation :)
I think the inner `<strong>` either needs to be escaped via i18n:name, or you place simple tag like `<tal:i18n i18n:translate="navigation_status">` only around the label “Status”.

src/euphorie/client/model.py
```diff
@@ -639,6 +640,12 @@ class SurveySession(BaseObject):
         default=None,
     )
 
+    completion_percentage = schema.Column(
+        types.Integer,
+        nullable=True,
+        default=0,
+    )
+
```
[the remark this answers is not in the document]
pysailor: I added such an upgrade step.
Independently, I can help you get your local alembic installation set up.

reinhardt:
> With every save, I need to wait at least an additional 1.5 seconds.

I was a bit worried about that.

I've been thinking about doing only one query that gets all the tree items and then process them in python code. Do you think that would solve the issue?

Btw, is there a method that actually constructs a tree from the "tree" table?

PR #246, opened 2020-03-17
WIP Action plan redesign

pysailor:
This PR does severval things:

1) Integrate the re-design of action plan regarding standard measures: instead of a pre-fill per measure, we have a single modal where many standard measured can be picked at once

2) The fields "action_plan" and "prevention_plan" get merged into a single new field "action". Both in the class content.Solution and in the table "action_plan". The old values, now redundant, are kept, in case we need them for refining the upgrade step per country. Cleaning up the old values will happen later

3) The "measures in place", formerly saved in a JSON structure, also get converted to action_plan items, so that we are better able to handle them together with the other measures. Again, the old values are kept (for now).

### DIFF
--- a/src/euphorie/client/browser/risk.py
+++ b/src/euphorie/client/browser/risk.py
@@ -472,9 +472,13 @@ class IdentificationView(BrowserView):
         }
         existing_measures = []
         for solution in self.solutions:
+            if self.italy_special:
+                text = solution.action
+            else:
+                text = solution.description
             existing_measures.append(
                 {
-                    "text": solution.action,
+                    "text": text,
                     "active": solution.id in saved_standard_measures,
                     "solution_id": solution.id,
                     "plan_type": "in_place_standard",
--- a/src/euphorie/client/browser/risk.py
+++ b/src/euphorie/client/browser/risk.py
@@ -835,10 +835,16 @@ class ActionPlanView(BrowserView):
             existing_measures = [
                 txt.strip() for (txt, active) in self.get_existing_measures() if active
             ]
+            self.active_standard_measures = {
+                str(getattr(measure, "solution_id", "")): measure
+                for measure in context.standard_measures
+            }
+
             solutions = []
             for solution in self.risk.values():
                 if not ISolution.providedBy(solution):
                     continue
+                solution_id = solution.id
                 description = (getattr(solution, "description", "") or "").strip()
                 prevention_plan = (
                     getattr(solution, "prevention_plan", "") or ""
@@ -853,7 +859,7 @@ class ActionPlanView(BrowserView):
                             "action_plan": solution.action_plan,
                             "prevention_plan": solution.prevention_plan,
                             "requirements": solution.requirements,
-                            "id": solution.id,
+                            "id": solution_id,
                         }
                     )
             self.solutions = solutions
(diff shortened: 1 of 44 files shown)


src/euphorie/client/browser/risk.py
```diff
@@ -447,8 +451,43 @@ def proceed_to_next(self, reply):
...
+                    "text": measure.action,
+                    "active": True,
+                    "solution_id": measure.id,
+                    "plan_type": "in_place_custom",
+                }
+            )
+        return existing_measures
+
+    @memoize
+    def get_existing_measures_old(self):
```
ale-rt: I do not see this called anywhere:
```
[ale@emily Euphorie]$ grep -R get_existing_measures_old
src/euphorie/client/browser/risk.py:    def get_existing_measures_old(self):
src/euphorie/client/browser/risk.py:    def get_existing_measures_old(self):
```

src/euphorie/client/browser/risk.py
```diff
@@ -634,7 +672,17 @@ def skip_evaluation(self):
...
     def get_existing_measures(self):
+        return [
+            measure
+            for measure in (
+                self.context.in_place_standard_measures
+                + self.context.in_place_custom_measures
+            )
+        ]
+
+    def get_existing_measures_old(self):
```
ale-rt: I do not see this called anywhere:
```
[ale@emily Euphorie]$ grep -R get_existing_measures_old
src/euphorie/client/browser/risk.py:    def get_existing_measures_old(self):
src/euphorie/client/browser/risk.py:    def get_existing_measures_old(self):
```

src/euphorie/client/browser/risk.py
```diff
@@ -634,7 +672,17 @@ def skip_evaluation(self):
             return True
         return False
 
+    @memoize
     def get_existing_measures(self):
+        return [
```
ale-rt: isn't the code in `get_existing_measures` equivalent to:
```
return list(self.context.in_place_standard_measures) + list(self.context.in_place_custom_measures)
```
?

src/euphorie/deployment/upgrade/alembic/versions/27_.py
```diff
@@ -0,0 +1,38 @@
...
+    op.execute("""
+UPDATE action_plan ap
+    SET "action" = CASE
+    WHEN ap.prevention_plan is not NULL THEN ap.action_plan || E'\n' || ap.prevention_plan
+    ELSE ap.action_plan END""")
+    op.alter_column('action_plan', 'plan_type', nullable=False)
+
+
+def downgrade():
+    op.drop_index(op.f('ix_action_plan_plan_type'), table_name='action_plan')
```
ale-rt: Is this correct, I do not see this mentioned elsewhere, but maybe it is some magic added by sqlalchemy

src/euphorie/client/browser/risk.py
```diff
@@ -447,8 +451,43 @@ def proceed_to_next(self, reply):
         )
         return self.request.response.redirect(url)
 
+    @property
+    @memoize
+    def solutions(self):
+        if not self.risk:
```
ale-rt: Given getattr is used a lot through the rest of the code I would go for the more concise:
```
return getattr(self.risk, "_solutions", [])
```

src/euphorie/client/browser/risk.py
pysailor: Yes, true. I kept it as a "backup", in case I needed to go back to fetching legacy data.
I will remove it once I have gotten feedback from several countries that they are happy with the result of the migration.

src/euphorie/client/browser/risk.py
pysailor: Yes, that is actually much simpler and easier to read, thanks

src/euphorie/deployment/upgrade/alembic/versions/27_.py
pysailor: Yes, it it mentioned a few lines above:
`op.create_index(op.f('ix_action_plan_plan_type'), 'action_plan', ['plan_type'], unique=False)`

src/euphorie/client/browser/risk.py
pysailor: makes sense, thanks!

PR #250, opened 2020-04-23
Remove unnecessary class

ale-rt:


pysailor:
It seems I missed this PR, but we talked about the issue, and it was fixed in a very similar fashion in #251

PR #269, opened 2020-12-02
Switch from travis to github actions

ale-rt:


ale-rt:
This also updates Plone from 5.1.5 to Plone 5.1.7 latest

PR #275, opened 2021-01-05
Run black flake8 and isort on the complete codebase

ale-rt:
Also add a pre-commit hook to run black flake8 and isort


pysailor:
For the sake of clarity, I will rename this PR like I did in NuPlone.
It does a lot more than just adding a pre-commit hook.

pysailor:
The previous run was just green, but I think you just pushed more.
I'll wait for the next test run to finish.
First tests in the browser look promising.

PR #279, opened 2021-01-08
Remove grok

ale-rt:


pysailor (commented):
Can we squash the 2 commits that both alter the coverage threshold?

And general question: do you think it is wise to enforce this tight coverage threshold for a build to pass? Does that mean that any new upgrade step (which will likely not have test coverage) will prevent a successful build?

ale-rt:
> Can we squash the 2 commits that both alter the coverage threshold?
> 

No, one of them is already on master.

> And general question: do you think it is wise to enforce this tight coverage threshold for a build to pass? Does that mean that any new upgrade step (which will likely not have test coverage) will prevent a successful build?

Yes, I would like to keep the coverage high. 
If you want to add uncovered code please add the `# pragma: no cover` comment.

PR #283, opened 2021-01-12
Provide bogus footer

pysailor:
In NuPlone the basic layout template defines a "tile" slot for footer.
Since `queryMultiAdapter((context, request), Interface, name="footer")` will find
the Plone footer portlet (which we don't want to display), we need to provide a bogus value
that will result in None


ale-rt:
There is something fishy here...
BTW what is `api.content.get_view("footer", context, request)` returning?

ale-rt:
Other questions:
1. Did I messed up the previous behavior? Why before the portlet was not fetched?
2. Does it make sense to move the registry record to NuPlone? Or should NuPlone provide a `footer` tile himself?
3. Why https://demo-latest-py3.plone.org/@@footer returns a 404

pysailor:
> BTW what is `api.content.get_view("footer", context, request)` returning?

<Products.Five.metaclass.SimpleViewClass from /home/thomas/.buildout/eggs/Products.CMFPlone-5.1.7-py2.7.egg/Products/CMFPlone/browser/templates/footer.pt object at 0x7f88cc0df450>

> Why before the portlet was not fetched?

Before, a different portlet was fetched for "footer". In this config, a "footer" tile is defined:
https://github.com/euphorie/Euphorie/blob/master/src/euphorie/deployment/config.ini#L22
As I learned recently, the app-config worked cumulatively. That means, even though in {buildout-dir}/etc/euphorie.ini no footer-tile was defined, the one from the config.ini linked above was already active. That means, in the "footer" slot, this tile was being shown: https://github.com/euphorie/NuPlone/blob/master/plonetheme/nuplone/tiles/configure.zcml#L15
Since that tile only contains JS code, there was nothing visible to the user.

Now, without a specific footer definition, the footer view from Plone is found, leading to a visible glitch:

Screenshot 2021-01-13 at 09 04 25

> Does it make sense to move the registry record to NuPlone? Or should NuPlone provide a footer tile himself?

Yes, I think it makes sense that NuPlone provides an empty default footer tile. Or, we could even remove the slot here
https://github.com/euphorie/NuPlone/blob/2.x/plonetheme/nuplone/skin/templates/layout.pt#L52

> Why https://demo-latest-py3.plone.org/@@footer returns a 404

I do not know. Maybe in Plone 5.2 it does not exists any more?

In plone 5.1 I get this. But maybe we can simply ignore this issue, if it goes away in plone 5.2?
Screenshot 2021-01-13 at 09 05 05

ale-rt:
Thanks!
My proposal is to wait a little bit until we reach a state where we are capable of running the site on Plone 5.2.
Then reevaluate and take some action in NuPlone.

pysailor:
ok, sounds like a good plan!

pysailor:
Closing as invalid

PR #284, opened 2021-01-12
Modernize import for getAllowedSizes

ale-rt:


ale-rt:
The code is identical:
- https://github.com/plone/Products.CMFPlone/blob/5.1.x/Products/CMFPlone/utils.py#L684
- https://github.com/plone/plone.app.imaging/blob/b6dc57a888ffa23936003cdd06c16c5301f1c72b/src/plone/app/imaging/utils.py#L12

PR #288, opened 2021-01-18
Remove more persistent stuff

pysailor:


ale-rt (approved):
LGTM, but next time please it would be good to not commit on an existing upgrade or to "touch" (i.e. rename the timestamp with a newer one) the existing upgrade folder.

pysailor:
Thanks, I will keep that in mind (renaming the upgrade). I can even do that now, unless you have already cherry-picked the commits?

ale-rt:
Feel free to rename whatever you like (note that the code is not black)

ale-rt (approved):

PR #297, opened 2021-03-12
Assorted fixed needed to generate a report

ale-rt:


src/euphorie/client/utils.py
```diff
@@ -148,7 +148,7 @@ def grow(tree, nodes):
                 tree[(node.path, node.type, node.id)] = {}
 
     def prune(tree):
-        for k in tree.keys():
+        for k in list(tree):
```
pysailor, line 151: Wow, I would love to hear why `tree.keys()` causes a problem.

PR #298, opened 2021-03-12
Use Plone 5.2.4

ale-rt:


versions-plone-5.2-latest.cfg
```diff
@@ -400,4 +401,80 @@ plone.app.standardtiles = 2.3.2
...
+markdown = 3.1.1
+mock = 3.0.5
+
+# BBB only packages (to be removed with Plone 5.3 or 6.0 and in 5.2 on Python 3.6+
+plone.app.controlpanel                = 4.0.1
+
+# Archetypes
+# part of core, listed separate b/c of planned removal for 6.0 and 5.2 on Python 3
+archetypes.referencebrowserwidget     = 2.5.11
+archetypes.schemaextender             = 3.0.1
```
pysailor, line 468: Do we really need this?

pysailor (approved):
LGTM, apart from some version pinnngs that I don't know if Euphorie really needs them. There's archetypes-related stuff, but also things like GitPython or pdbpp

versions-plone-5.2-latest.cfg
ale-rt: it is automatically generated...

pysailor: From what / which script? `show-picked-versions = true`?
That would mean that Euphorie pulls in those dependencies.
Or maybe I'm not getting something

ale-rt: Nevermind, I would not focus on those pins, really...

pysailor: yeah, fine for me.
I was just wondering where this comes from, as we generally try to avoid redundancy and repeating things that we actually do not need.

PR #300, opened 2021-03-15
Use wkhtmltopdf to render PDF files

ale-rt:


pysailor (changes requested):
Very nice! Looking pretty good in general already

Apart from differences in font size, which I think we can live with for now, I see 2 issues that need to be addressed.

1) Missing page numbers.
Look at any report created with smartprint. Bottom right (footer) we have "Page X of TOTAL"

Screenshot 2021-03-15 at 13 42 39

2) Page breaks don't respect blocks
In the smartprint-generated PDF, there are several control statements in place that tell the PDF generator when a page break is allowed.
In the wkhtml version, it seems those checks are not in place. This can lead to some very ugly results:
Screenshot 2021-03-15 at 13 42 20

ale-rt:
I added a page summary here: https://github.com/euphorie/Euphorie/pull/300/commits/a17a6b199633585f2b7614547c0531356cef4ac1
Note you have to use the compiled version of wkhtmltopdf for it to work properly: https://wkhtmltopdf.org/downloads.html

Also I fixed the css by adding:
```
* {
  overflow: visible !important;
}
```
But you might want to adapt the css further.
Also I do not see any meta charset there...

pysailor (changes requested):
It's getting better.

Still, some problems remain.

1) Missing footer translation
See https://github.com/euphorie/Euphorie/blob/12.x/src/euphorie/client/browser/templates/plain.pt#L33 and https://github.com/euphorie/Euphorie/blob/12.x/src/euphorie/client/browser/session.py#L935
Original:
pdf-orig

2) Footer text out of proportion
Compared to the original (and the rest of the text), the footer text is too large

3) Encoding issues
Non-ascii text seems to break
pdf-wkhtml

4) The bar, if it does not start with "grey" apparent is missing the rounded corner

You probably need to compare a number of examples in our test servers in different combinations (language, grade of filling). Let me know if you need concrete example URLs.

ale-rt:
Well, the translation is an easy fix.
The rest was already anticipated:
- the encoding issues are caused by the missing meta tag
- the css should be adapted

I will see what I can do.

ale-rt:
Check this out please

pysailor:
Much better! I also tweaked the CSS a bit more, it seems good enough now.
The only thing that I was not able to figure out: for some reason, there always seems to be an extra blank page at the end. We can fix that later. I will this as soon as all tests are green again.

PR #302, opened 2021-03-18
Attempt to fix SQL query

ale-rt:
Should fix:
```
Can't determine which FROM clause to join from
```


pysailor (changes requested):
Only works as long as there is no content to show ;-)
Go to action plan, add a measure and set the "Planning start" to any day in the current or coming 2 months.

pysailor:
Superseded by #303

PR #303, opened 2021-03-24
Fix SQL query with multiple joins

pysailor:
In sqlalchemy 1.3, join() is more strict, see
https://docs.sqlalchemy.org/en/14/changelog/migration_13.html#key-behavioral-changes-orm
Therefore, define which table to select from (and simplify joining of Risk table by using
the parent-relationship)


ale-rt (approved):
Sweet!

PR #311, opened 2021-05-19
Add a new view @@get-resources-timestamp

ale-rt:
A new view ``@@get-resources-timestamp`` has been introduced to not use
the protected view ``@@refresh-resources-timestamp`` which will be
forbidden by the latest security hotfix


ale-rt:
This is not strictly needed anymore after with version 1.2 of the hotfix but it seems a good idea to merge this anyway.

PR #334, opened 2021-11-08
Sniff the user agent and issue a warning if we detect IE

pysailor:
Also, prevent loading of our main JavaScript bundle - with an outdated browser, we simply do not offer any service at all any more.
Refs #SCR-1524
Implements https://github.com/syslabcom/oira.prototype/issues/302


ale-rt (approved):
LGTM

ale-rt (approved):

PR #335, opened 2021-11-18
Prevent XSS in text entered by the user (description, measure text).

pysailor:
Clean texts that appear in a structure or pat-markdown context using the safe_html transform


ale-rt (approved):
LGTM

PR #336, opened 2021-12-01
Keep country after registration

reinhardt:
Login/Registration: Use webhelpers to get country URL. Fixes redirect to default country when came_from is missing.

Refs [SCR-1586](https://jira.syslab.com/browse/SCR-1586)


pysailor:
> I get the same failures on the 12.x-portal-renewal branch, so they don't seem to be related to my changes.

Yes, I think especially the (very very old) doctests don't run any more under python3. Needs looking into

pysailor (approved):
LGMT

PR #337, opened 2021-12-01
Translation of validation messages

reinhardt:
Validation of the email field depens on https://github.com/euphorie/NuPlone/pull/33

Refs SCR-1587


pysailor (approved):
I like this elegant solution!

PR #338, opened 2021-12-02
Warning about changed structure

reinhardt:
Show a warning when republishing a tool with changed structure.

Note: Added modules/risks are not considered a significant change. Only removed items are warned about.

Refs SCR-1600


pysailor (approved):
Simple and elegant

PR #340, opened 2021-12-15
[WIP] portal renewal test fixes

thet:


src/euphorie/client/browser/configure.zcml
```diff
@@ -571,6 +571,14 @@
...
       layer="euphorie.client.interfaces.IClientSkinLayer"
       />
+  <browser:page
+      name="register"
+      for="*"
+      permission="zope.Public"
+      class=".login.Login"
+      template="templates/login.pt"
+      layer="euphorie.client.interfaces.IClientSkinLayer"
+      />
```
pysailor, line 581: Hi @thet, why is it necessary to have this definition?
It does not seem correct to me. If we don't need this for normal usage, why do we need it in the tests?

pysailor: I see now, it was how you fixed client: test_functional_login.RegisterTests
I will try something different: Since the whole login / register is all one view now, I will try to just call the login view instead.

PR #341, opened 2021-12-16
Added a toggle (registry entry) that allows to disable self-registration

pysailor:


src/euphorie/client/browser/login.py
```diff
@@ -327,6 +327,11 @@ def email_message(self):
 
     def __call__(self):
         self.errors = {}
+        allowed = api.portal.get_registry_record(
+            "euphorie.allow_self_registration", default=True
```
ale-rt, line 331: Note that Plone already has a `plone.enable_self_reg`

pysailor:
Replaced by #342

PR #342, opened 2021-12-17
Respect security setting for self-registration

pysailor:
Make self-registration the default - but disable the possibility to register if that setting is turned off.

### DIFF
--- a/src/euphorie/deployment/profiles/default/registry.xml
+++ b/src/euphorie/deployment/profiles/default/registry.xml
@@ -311,4 +311,13 @@
     </field>
     <value purge="false">--margin-bottom 1cm --margin-left 1.5cm --margin-right 1.5cm --margin-top 2cm --disable-javascript --viewport-size 10000x10000</value>
   </record>
+  <record name="plone.enable_self_reg" interface="Products.CMFPlone.interfaces.controlpanel.ISecuritySchema" field="enable_self_reg">
+    <field type="plone.registry.field.Bool">
+      <default>False</default>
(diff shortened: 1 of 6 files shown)


src/euphorie/deployment/profiles/default/registry.xml
```diff
@@ -311,4 +311,13 @@
     </field>
     <value purge="false">--margin-bottom 1cm --margin-left 1.5cm --margin-right 1.5cm --margin-top 2cm --disable-javascript --viewport-size 10000x10000</value>
   </record>
+  <record name="plone.enable_self_reg" interface="Products.CMFPlone.interfaces.controlpanel.ISecuritySchema" field="enable_self_reg">
+    <field type="plone.registry.field.Bool">
```
ale-rt, line 315: The field element can be omitted if I am not wrong, given that Plone should already have it.

src/euphorie/upgrade/deployment/v1/20211216172426_toggle_for_self_registration/registry.xml
```diff
@@ -0,0 +1,11 @@
+<registry xmlns:i18n="http://xml.zope.org/namespaces/i18n">
+  <record name="plone.enable_self_reg" interface="Products.CMFPlone.interfaces.controlpanel.ISecuritySchema" field="enable_self_reg">
+    <field type="plone.registry.field.Bool">
```
ale-rt, line 3: The field element can be omitted if I am not wrong, given that Plone should already have it.

ale-rt (approved):

PR #349, opened 2022-02-01
Fix data import

### LINKED ISSUE #348: Partially revert the change done to handle huge XML files
This change https://github.com/euphorie/Euphorie/commit/a0e115dc8c68227dd373f05f103155dcef269b74 broke the import of files with Cyrillic characters because reparsing.

The reason is that the characters are converted in to entities and when the xml is reparsed after it is converted to a string we have this error:
```
-> sector = lxml.objectify.fromstring(lxml.etree.tostring(sector_tree))
(Pdb) n
lxml.etree.XMLSyntaxError: PCDATA invalid Char value 16, line 59, column 1965
```

ale-rt:
Closes #348

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,7 +4,8 @@ Changelog
 13.0.8 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Fix data import (see #348)
+  [ale-rt]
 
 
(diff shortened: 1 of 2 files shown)


pysailor (approved):
Nice one, thanks!

PR #350, opened 2022-02-09
Upgrade bundle name

thet:
Since Patternslib 6.4.0 we have single-name bundles.
Minified and unminified, both are named bundle.min.js.
This way we do not need to change the resource registry when developing JavaScript but only recompile the bundle with webpack in development mode.
The rename can also be justified as that the development bundle is already Babel-transpiled and Webpack-compiled, so a bundle.min.js makes sense naming-wise also.

This change removes the ``bundle.js`` name when starting in development mode.


src/euphorie/client/resources/oira/help/bg/illustrations/2014-01-02-work-environment-site-related.html
```diff
@@ -13,12 +13,12 @@
 		<link
 			id="all-styles"
 			rel="stylesheet"
-			href="../../../style/all.css"
+			href="/assets/oira/style/all.css"
 			type="text/css">
```
pysailor, line 17: @thet Something has gone wrong here. Your PR will introduce a broken path in all the Help files.

Any paths that start with "/assets" need to be made relative or replaced with "++resource++euphorie.resources/". See the script https://github.com/euphorie/Euphorie/blob/main/scripts/proto2diazo.py

pysailor (changes requested):
Your change probably makes sense. But all the Help files are broken.

We can debug this together if you want. Usually, the `proto2diazo` script takes care of creating correct paths (called via `make resources_install`). Maybe something in the general structure has changed, so that the script needs to be adjusted.

thet:
Obsolete via: https://github.com/euphorie/Euphorie/pull/358

PR #353, opened 2022-02-16
Remove unused pat-validation option type from date fields in forms.

thet:


pysailor:
I believe this need a current bundle?
I will re-issue this PR against a branch

thet:
It's not necessarily needed but it removes a warning about a unknown option ``type`` which was removed in Patternslib 7.0.

PR #359, opened 2022-03-14
Security fix: transfer user's sessions

pysailor:
When the session(s) of a guest account are transferred to an existing regular account, make sure that the source account really is currently logged in. Do not allow to transfer sessions from any other account (guest or otherwise) apart from the currently active guest account.

### DIFF
--- a/src/euphorie/client/browser/login.py
+++ b/src/euphorie/client/browser/login.py
@@ -18,9 +18,11 @@ from euphorie.client.model import get_current_account
 from euphorie.content.survey import ISurvey
 from plone import api
 from plone.memoize.view import memoize
+from plone.session import tktauth
 from plone.session.plugins.session import cookie_expiration_date
 from plonetheme.nuplone.tiles.analytics import trigger_extra_pageview
 from Products.CMFCore.utils import getToolByName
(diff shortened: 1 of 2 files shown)


src/euphorie/client/browser/login.py
```diff
@@ -89,9 +92,23 @@ def login(self, account, remember):
...
+        if not cookie:
+            return
+        ticket = decodebytes(safe_bytes(cookie))
+        try:
+            cookie_values = tktauth.splitTicket(ticket)
+        except ValueError:
+            return
+        if len(cookie_values) < 2 or cookie_values[1] != account_id:
+            return
         account = get_current_account()
```
ale-rt, line 112: I am missing some details here, but that does not seem to be a valid fix...
Isn't it fixed by removing all the cruft and doing:
```
model.SurveySession.account_id == account.id
```
I am probably missing quite a lot of information...

ale-rt (approved):

src/euphorie/client/browser/login.py
ale-rt: Never mind I think I got it... but still I do not understand why we are not using `account.id`

PR #362, opened 2022-03-16
Do not use the guest_account_id in the login and register forms

ale-rt:
It is redundant and we have better ways to know it

### DIFF
--- a/src/euphorie/client/browser/login.py
+++ b/src/euphorie/client/browser/login.py
@@ -88,16 +88,44 @@ class Login(BrowserView):
                 120 * 24 * 60 * 60
             )  # noqa: E501
 
-    def transferGuestSession(self, account_id):
-        """Transfer session(s) from guest account to an existing user account"""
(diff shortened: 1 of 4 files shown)


src/euphorie/client/browser/login.py
```diff
@@ -88,16 +88,44 @@ def login(self, account, remember):
...
+            # We came here as anonymous user, no need to proceed
             return
-        account = get_current_account()
+
+        old_authenticated_account_id = old_credentials[0]
+
+        if (
+            Session.query(model.Account)
+            .filter(
+                model.Account.account_type == config.CONVERTED_ACCOUNT,
```
ale-rt, line 112: I am not sure about this line, maybe it should be:
```suggestion
                model.Account.account_type == config.GUEST_ACCOUNT,
```
It might also be that the method should not even exist.

pysailor (approved):
LGTM

PR #364, opened 2022-03-17
Exception Handling: Fallback

reinhardt:
Fall back to a basic error template if the regular one fails to render. This avoids falling back to a zExceptions view which is XSS vulnerable.

Borrowed from the CMFPlone ExceptionView.

Refs syslabcom/scrum#158


pysailor (approved):
LGTM

PR #366, opened 2022-03-17
Password reset: Fix email tampering

reinhardt:
Password reset: Only use X-Forwarded-For if it contains an IP address. Fixes email tampering attack.

Refs syslabcom/scrum#158


pysailor (approved):
LGTM

PR #369, opened 2022-03-18
CSRF for SQL writes

reinhardt:
Make CSRF protection kick in if SQL data is being written.

This is a little hacky because `safeWrite` doesn't work for the objects that are added (they have no _p_oid).

The tests do pass but there could of course still be some false positives of the CSRF protection. I would only expect a `@@confirm-action` redirect in this case, but maybe it pays off to do some more thorough manual testing.

Refs syslabcom/scrum#158


reinhardt:
The tests do pass but there could of course still be some false positives of the CSRF protection. I would only expect a `@@confirm-action` redirect in this case, but maybe it pays off to do some more thorough manual testing.

PR #371, opened 2022-03-23
XML export enhancement

pysailor:
When a tool editor wants to export an OiRA tool as XML, they now have the option to de-select certain element, such as images or legal texts, so that the tool export is useful for translation.
Additionally, a text-only version can be included in the export, so that a simple word count will be possible. This is useful for translators.

See https://github.com/syslabcom/scrum/issues/147


ale-rt (approved):
I barely understand what is going on but looks technically very well done

reinhardt (approved):

pysailor:
Thanks, that's valuable feedback. Since this XML export is only available to CMS users, I would say that it's safe to merge this PR.

PR #375, opened 2022-03-29
Consolidate previews

reinhardt:
Show descriptions in print preview and measures in Word preview.

Refs syslabcom/scrum#146

### DIFF
--- a/src/euphorie/client/browser/templates/panel-contents-preview.pt
+++ b/src/euphorie/client/browser/templates/panel-contents-preview.pt
@@ -63,11 +63,11 @@
                             "
                         >
                           <em>${node/title}</em>
-                          <p tal:replace="structure description"></p>
+                          <p tal:replace="structure python:webhelpers.get_safe_html(description)"></p>
                           <ul class="manual-checklist"
                               tal:condition="solutions"
                           >
-                            <li tal:repeat="solution solutions"><tal:solution replace="structure solution">text</tal:solution></li>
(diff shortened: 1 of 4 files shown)


src/euphorie/client/browser/templates/panel-contents-preview.pt
```diff
@@ -57,11 +57,13 @@
...
                         <li class="risk"
                             tal:define="
+                              description python:view.get_description(node);
                               legal_references python:view.get_legal_references(node);
                               solutions python:view.get_solutions(node);
                             "
                         >
-                            ${node/title}
+                          <em>${node/title}</em>
+                          <p tal:replace="structure description"></p>
```
pysailor, line 66: Displaying content that can be entered by a user (either end user or tool creator) directly as rendered HTML is _verboten_ because of potential XSS attacks. See e.g. e9746b7023d55bb126a8aa484dc56fb3a440beff
```suggestion
                          <p tal:replace="structure python:webhelpers.get_safe_html(description)"></p>
```

Please also apply the same `get_safe_html`  check to the other places where `structure` is used (for **solution** and **legal_references** below).

pysailor (changes requested):
I like this creative solution, even if it does not strictly follow proto.

Please have a look at my findings. The XSS protection is not strictly in the scope of the original ticket, but I was looking at this feature with a test session that has nasty HTML with Javascript all over the place - and the remedy is quite simple.

pysailor (approved):
LGTM!

PR #381, opened 2022-04-13
Training questions: Randomize

reinhardt:
Training questions: Pick a random subset of questions per training session.

Refs syslabcom/scrum#195

### DIFF
--- a/src/euphorie/client/browser/training.py
+++ b/src/euphorie/client/browser/training.py
@@ -245,7 +245,7 @@ class TrainingView(BrowserView, survey._StatusHelper):
         """
         survey = self.webhelpers._survey
         all_questions = api.content.find(
-            obj=survey,
+            context=survey,
             portal_type="euphorie.training_question",
         )
         num_training_questions = getattr(survey, "num_training_questions", None) or len(
--- a/src/euphorie/client/browser/training.py
+++ b/src/euphorie/client/browser/training.py
@@ -309,9 +309,9 @@ class TrainingView(BrowserView, survey._StatusHelper):
     @view_memoize
     def questions(self):
         training = self.get_or_create_training()
-        question_ids = loads(training.answers).keys()
+        answer_history = loads(training.answers)
         survey = self.webhelpers._survey
-        return [survey.get(q_id) for q_id in question_ids]
+        return [survey.get(q_id) for q_id in answer_history]
 
     @property
     def enable_training_questions(self):
--- a/src/euphorie/client/browser/training.py
+++ b/src/euphorie/client/browser/training.py
@@ -245,10 +245,8 @@ class TrainingView(BrowserView, survey._StatusHelper):
         """
         survey = self.webhelpers._survey
         all_questions = api.content.find(
-            **{
-                "portal_type": "euphorie.training_question",
-                "path": "/".join(survey.getPhysicalPath()),
-            }
+            obj=survey,
+            portal_type="euphorie.training_question",
         )
         num_training_questions = getattr(survey, "num_training_questions", None) or len(
             all_questions
--- a/src/euphorie/client/browser/training.py
+++ b/src/euphorie/client/browser/training.py
@@ -244,14 +244,17 @@ class TrainingView(BrowserView, survey._StatusHelper):
         None.
         """
         survey = self.webhelpers._survey
-        all_questions = survey.listFolderContents(
-            {"portal_type": "euphorie.training_question"}
+        all_questions = api.content.find(
+            **{
+                "portal_type": "euphorie.training_question",
+                "path": "/".join(survey.getPhysicalPath()),
+            }
         )
         num_training_questions = getattr(survey, "num_training_questions", None) or len(
             all_questions
         )
-        questions = sample(all_questions, k=num_training_questions)
-        return {q.getId(): None for q in questions}
+        questions = sample(list(all_questions), k=num_training_questions)
+        return {q.getId: None for q in questions}
 
     @memoize
     def get_or_create_training(self):
@@ -293,8 +296,11 @@ class TrainingView(BrowserView, survey._StatusHelper):
         if not getattr(survey, "enable_web_training", False):
             return ""
         view_name = "slide_question_success"
-        if survey.listFolderContents(
-            {"portal_type": "euphorie.training_question"}
+        if api.content.find(
+            **{
+                "portal_type": "euphorie.training_question",
+                "path": "/".join(survey.getPhysicalPath()),
+            }
         ) and self.training_status not in ("correct", "success"):
             view_name = "slide_question_intro"
         return "{}/@@{}".format(self.context.absolute_url(), view_name)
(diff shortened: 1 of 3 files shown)


src/euphorie/client/browser/training.py
```diff
@@ -238,6 +239,20 @@ def session(self):
...
+        """
+        survey = self.webhelpers._survey
+        all_questions = survey.listFolderContents(
+            {"portal_type": "euphorie.training_question"}
+        )
+        num_training_questions = getattr(survey, "num_training_questions", None) or len(
+            all_questions
+        )
+        questions = sample(all_questions, k=num_training_questions)
+        return {q.getId(): None for q in questions}
```
ale-rt, line 254: Isn't it better to get the brains from the catalog instead of the objects?

src/euphorie/client/browser/training.py
```diff
@@ -285,8 +302,10 @@ def question_intro_url(self):
     @property
     @view_memoize
     def questions(self):
+        training = self.get_or_create_training()
+        question_ids = loads(training.answers).keys()
```
ale-rt, line 306: `.keys()` looks like it is not needed

src/euphorie/client/browser/training.py
```diff
@@ -285,8 +302,10 @@ def question_intro_url(self):
...
         survey = self.webhelpers._survey
-        return survey.listFolderContents({"portal_type": "euphorie.training_question"})
+        return [survey.get(q_id) for q_id in question_ids]
```
ale-rt, line 308: It seems to me you are getting the objects then you are using their id, dump them in to a json and then load the json to get the id and fetch the objects once again.
It seems that this story can be simplified a little.

src/euphorie/client/browser/training.py
reinhardt: Quicker, you mean? Yes, possibly.

src/euphorie/client/browser/training.py
reinhardt: I'm not sure I understand you correctly. I do get the objects with `survey.get(q_id)`, but they are not dumped to json. `questions()` returns a list of objects.

Or do you mean that `get_or_create_training()` fetches the objects? It only does so once when the training is initialized, and I'm changing that to use the catalog, see your other comment.

I could go through all the uses of `self.questions` and check whether they really need the objects, but I wonder whether it's worth the effort.

src/euphorie/client/browser/training.py
```diff
@@ -238,6 +239,23 @@ def session(self):
...
+    def get_initial_answers(self):
+        """Pick a subset of questions, shuffle them, and initialize the answers with
+        None.
+        """
+        survey = self.webhelpers._survey
+        all_questions = api.content.find(
+            **{
+                "portal_type": "euphorie.training_question",
+                "path": "/".join(survey.getPhysicalPath()),
+            }
```
ale-rt, line 251:
```suggestion
            obj=survey,
            portal_type="euphorie.training_question",
```

reinhardt: I wasn't aware of the `obj=` shorthand. That's handy, cheers.

src/euphorie/client/browser/training.py
reinhardt: I did straighten it out a little. See afb0b43

reinhardt:
@ale-rt, would you like to check again or will I merge?

src/euphorie/client/browser/training.py
```diff
@@ -238,6 +239,21 @@ def session(self):
         """Return the session for this context/request"""
         return self.context.session
 
+    def get_initial_answers(self):
+        """Pick a subset of questions, shuffle them, and initialize the answers with
+        None.
+        """
+        survey = self.webhelpers._survey
+        all_questions = api.content.find(
+            obj=survey,
```
ale-rt, line 248:
```suggestion
            context=survey,
```
Sorry I tricked you but the correct parameter is `context` and not `obj`: https://github.com/plone/plone.api/blob/e5f5d455a6cf8f87877c13425eff16dd41f217bf/src/plone/api/content.py#L591
:man_facepalming:

src/euphorie/client/browser/training.py
```diff
@@ -276,17 +294,19 @@ def question_intro_url(self):
         if not getattr(survey, "enable_web_training", False):
             return ""
         view_name = "slide_question_success"
-        if survey.listFolderContents(
-            {"portal_type": "euphorie.training_question"}
+        if api.content.find(
+            obj=survey,
```
ale-rt, line 298:
```suggestion
            context=survey,
```

src/euphorie/client/browser/training.py
```diff
@@ -276,17 +294,19 @@ def question_intro_url(self):
...
         return "{}/@@{}".format(self.context.absolute_url(), view_name)
 
     @property
     @view_memoize
     def questions(self):
-        survey = self.webhelpers._survey
-        return survey.listFolderContents({"portal_type": "euphorie.training_question"})
+        training = self.get_or_create_training()
+        answer_history = loads(training.answers)
+        return list(answer_history)
```
ale-rt, line 309: This is changed to return the question ids, right?
Then this bubbles up into a lot of changes in the code that otherwise could be avoided.

What about having a `question_ids` property that returns the question ids and restore the return type of the `questions` property?
Would this help in having a smaller patch?

reinhardt: But the point was to change the parts in the code that were getting the objects without needing them. Most calls of `questions()` actually don't need the objects. I can of course change it back but then your concern is not addressed.

ale-rt: I expressed myself poorly then.

What I am trying to achieve is to have both the properties `questions` and `question_ids`, in that way:
- `self.question_ids` will be used the many times you just need the object ids
- `self.questions` will be used the few time you need the objects and it will be

Does it make more sense now?

reinhardt: Yes, but why? There is one or maybe two cases where we need the objects. Why do we need a method for that?

reinhardt: In other words, doing it like that will not make the patch smaller.

ale-rt: There are many reasons and I would not like to waste lot of time with that.

The premises:
1. This PR already renames:
   - `previous_question` -> `previous_question_id`
   - `next_question` -> `next_question_id`
2. the code already has a `question_id` attribute and a `question` property.

So "the main because" is that as a casual "code reader/debugger" I would not expect `self.questions` to return ids, but rather objects.

This reason alone might already justify to rename the method `questions` to `question_ids`.
I think you agree with me that it would increase the consistency of the code, which is a desirable thing.
Keeping a `questions` method that returns the object might be handy but I am not sure it is needed.

Also I do not know is all this changes will break some other projects that depend on this package.

It might also be a good thing to start adding some deprecation warnings instead of just removing renaming methods.

reinhardt: I'm happy to rename `questions` to `question_ids`. The training feature is completely new and not even merged yet, so there can't be any code that depends on this. Therefore, and because you originally remarked that we should not get the objects if we don't need them, I wouldn't bother keeping a method that gets the objects.

ale-rt (approved):

PR #384, opened 2022-05-04
Disable pat-validation due to a bug in Chrome with date validation.

thet:


thet:
Not needed in Patternslib 7.10.0.

PR #393, opened 2022-05-24
Training toggle on country level

reinhardt:
The three levels of switches (global, country, tool) are potentially confusing and at least the wording should be improved, but we don't have time for that right now.

Refs syslabcom/scrum#295

### DIFF
--- a/src/euphorie/client/browser/webhelpers.py
+++ b/src/euphorie/client/browser/webhelpers.py
@@ -165,9 +165,11 @@ class WebHelpers(BrowserView):
     @property
     @memoize
     def use_training_module(self):
-        return api.portal.get_registry_record(
+        globally_enabled = api.portal.get_registry_record(
             "euphorie.use_training_module", default=False
         )
+        country_enabled = self.content_country_obj.enable_web_training
+        return globally_enabled and country_enabled
 
     @property
     @memoize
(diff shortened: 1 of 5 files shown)


src/euphorie/client/browser/webhelpers.py
```diff
@@ -165,9 +165,11 @@ def use_involve_phase(self):
     @property
     @memoize
     def use_training_module(self):
-        return api.portal.get_registry_record(
+        globally_enabled = api.portal.get_registry_record(
             "euphorie.use_training_module", default=False
         )
+        country_enabled = self.content_country_obj.enable_web_training
```
ale-rt, line 171: It is useless to get the `country_enabled` if globally_enabled is False.

src/euphorie/content/country.py
```diff
@@ -99,6 +99,17 @@ class ICountry(model.Schema, IBasic):
...
+    enable_web_training = schema.Bool(
+        title=_("label_enable_web_training", default="Enable Web Based Training?"),
+        description=_(
+            "help_enable_web_training_country",
+            default="If this option is activated, an online training can be enabled on "
+            "OiRA tools in this country.",
+        ),
+        required=False,
+        default=True,
+    )
```
ale-rt, line 111: It feels weird that the default for `euphorie.use_training_module` is `False` and this is `True`.

ale-rt (approved):
This is okish

src/euphorie/content/country.py
reinhardt: Yes, but they also do different things, which is worse, but no time to get it straight now.

src/euphorie/client/browser/webhelpers.py
reinhardt: I'll change that, thx.

PR #395, opened 2022-05-30
Fix translation of confirmation/validation messages

reinhardt:
Action plan: Defer translations until language negotiation is finished.

Refs syslabcom/scrum#251


ale-rt (approved):
Thanks for removing those instance attributes set in the call, I always see that as an antipattern!

PR #410, opened 2022-07-04
Completion Percentage Performance

reinhardt:
This reduces the number of postgresql queries per call of the view. We should still look into calling the view less often, though.

Refs syslabcom/scrum#285


ale-rt:
Another approach #419

ale-rt:
I am going to close this because it was solved in a different way in #419

PR #411, opened 2022-07-04
Remove the spooky ghost module

ale-rt:
Remove the ghost module and its tests because it is not used since quite some time and it is spooky.


ale-rt:
I added the hold merge badge on this PR but I do not thing there is any risk involved. Anyway I am test running this on my Euphorie installations to check for side effects.

Luckily the `PathGhost` objects should not be persistent, so there should be no risk in removing this obsolete class.

That would have been double spooky.

PR #412, opened 2022-07-05
map_images improvements

reinhardt:
* Bail out early if tool already has an image.
* Improved exception handling.
* Readability.

Refs syslabcom/scrum#349


reinhardt:
Replaced by #417 and https://github.com/EU-OSHA/oira.application.buildout/pull/14

PR #413, opened 2022-07-05
Update proto resources

reinhardt:
Updated to latest version of proto (19710d4dcd).

Fixes syslabcom/scrum#443


ale-rt:
It seems this contains files we do not want... especially media files

PR #416, opened 2022-07-06
Align flake8 with black

reinhardt:
See https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,7 @@ repos:
     rev: 3.9.2
     hooks:
       - id: flake8
+        args: [--max-line-length=88, --extend-ignore=E203]
   - repo: https://github.com/PyCQA/bandit
     rev: 1.7.4
     hooks:


.pre-commit-config.yaml
```diff
@@ -11,6 +11,7 @@ repos:
     rev: 3.9.2
     hooks:
       - id: flake8
+        args: [--max-line-length=88, --extend-ignore=E203]
```
ale-rt, line 14: AFAIK that has to be set in the `setup.cfg` or in the `pyptoject.toml`.

I am a bit surprised you need this.
It works for me like it is now :)
```
ale@code:~/Code/plone/projects/eu5rie/src/Euphorie$ pre-commit run flake8 --all
flake8...................................................................Passed
```

reinhardt: Ah, now it makes sense. This first came up when I reused the pre-commit config for oira.application.buildout. This change is actually not needed if I update one of the files you mentioned in the other repo.

PR #418, opened 2022-07-06
Fix the way get_survey_templates handles a survey with an unset category

ale-rt:
Assuming that `categories == None`, the second `if` in:
```
                if not isinstance(categories, list):		
                    categories = [categories]		
                if not categories:		                if not categories:
                    categories = ["None"]
```
was never entered.

That caused `categories` to be modified in the end to `[None]`.

Then sorting:
```
return sorted(survey_items, key=lambda x: (x[0], x[1].title))
```
would fail with:
```
TypeError: '<' not supported between instances of 'str' and 'NoneType
```

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,7 +4,7 @@ Changelog
 14.1.6 (unreleased)
 -------------------
 
-- Nothing changed yet.
+- Fix the way ``get_survey_templates`` handles a survey with an unset category.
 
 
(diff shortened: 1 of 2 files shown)


reinhardt (commented):
Seems sensible but tests are failing.

ale-rt:
Now the tests are green. From your comments I suppose you are fine having this merged

PR #419, opened 2022-07-06
Deprecate the @@update-completion-percentage view

ale-rt:
The view was called through pat-inject calls and was doing write on
read writing the completed_percentage column in the session table.

Now the column has been removed and the value is computed on the fly by
optimized code that puts less pressure on the SQL server.


ale-rt:
Let's merge this for the next round of deployments.
A review is anyway welcome.

reinhardt:
> The view was called through pat-inject calls and was doing write on
> read writing the completed_percentage column in the session table.

Just to clarify, it was not a write-on-read in the sense of an accidental write operation while intending to only read data. For one thing it didn't happen when rendering the risk view but in a separate request (via inject autoload). For another the write was deliberately done like this to avoid having to compute the value on-the-fly. Doing the latter was deemed too expensive because at that point we hadn't figured out how to do the calculation efficiently (without a lot of database queries).

reinhardt (approved):

ale-rt:
> > The view was called through pat-inject calls and was doing write on
> > read writing the completed_percentage column in the session table.
> 
> Just to clarify, it was not a write-on-read in the sense of an accidental write operation while intending to only read data. For one thing it didn't happen when rendering the risk view but in a separate request (via inject autoload). For another the write was deliberately done like this to avoid having to compute the value on-the-fly. Doing the latter was deemed too expensive because at that point we hadn't figured out how to do the calculation efficiently (without a lot of database queries).

I know all of that, I was the one removing:
```
            <a class="pat-inject"
               id="percentage-placeholder"
               hidden="hidden"
               href="${here/webhelpers/traversed_session/absolute_url}/@@update-completion-percentage"
               data-pat-inject="trigger: autoload; source: #document-content; target: #percentage-placeholder::element; loading-class:; history: none"
            >Compute completion percentage</a>
```
That snippet is modifying a row in the DB here https://github.com/euphorie/Euphorie/blob/b1caf09f4dc9463b6a3b2cebd63da2a2f5af2d23/src/euphorie/client/browser/webhelpers.py#L310

If you do not want to call it a write on read I am fine and totally not interested to any further debate on this topic.
What happened happened.

I still keep the old merge label here even if the PR was approved and @pilz [wrote he is in favour of merging this now](https://github.com/syslabcom/scrum/issues/285#issuecomment-1176017243) because next week I am on vacation and I cannot support you in case of issues.

So @reinhardt, I would say that given that you were the original author of the feature and you will be probably the one asked to firefight possible issues, it is up to you to decide if this should be merged now.

reinhardt:
The important thing is to know what happened previously to have a clear understanding of the changes.

Thanks for the great work optimizing this! I'll double check with @angeldasangel about merging it.

reinhardt:
> If we go for this we should merge #410

You probably mean if we **don't** go for this?

PR #422, opened 2022-07-15
assessments_list macro: Allow showing/hiding the lock separately.

reinhardt:
Refs syslabcom/scrum#474


pilz:
Put the hold label because the customer comes with more ideas (meeeeh)

PR #427, opened 2022-08-01
Added a way to serialize a session

ale-rt:
Added an `@@export.json` view that allows to export a session and the
related content in a single json file


ale-rt:
Adding hold merge because I would like to do some more tests.

ale-rt:
It seems to me it works pretty well now, removing the hold merge

reinhardt (approved):
Pretty cool.

PR #430, opened 2022-08-05
Login cookie tweaks

reinhardt:
* Removed the “Keep me logged in”-checkbox.
* Login cookie is always valid for 12 hours.
* Login times out after 30 minutes of inactivity.

Refs syslabcom/scrum#454


reinhardt:
Will be moved to customer package

PR #432, opened 2022-08-10
Convert deleted measures

reinhardt:
Change type of measure in session if it is deleted in ZODB so that it stays visible in the client.

Refs syslabcom/scrum#351

### DIFF
--- a/src/euphorie/client/browser/session.py
+++ b/src/euphorie/client/browser/session.py
@@ -329,8 +329,15 @@ class Profile(SessionMixin, AutoExtensibleForm, EditForm):
             # Touch means: the modification timestamp is set.
             # But we need to make sure the refreshed marker is up to date!
             survey_session.refresh_survey(survey)
-            return survey_session
+            new_session = survey_session
+        else:
+            new_session = self.rebuild_session(survey, profile)
 
+        new_session.update_measure_types(survey)
+        return new_session
+
+    def rebuild_session(self, survey, profile):
+        survey_session = self.session
         params = {}
         # Some values might not be present, depending on the type of survey session
         _marker = object()
--- a/src/euphorie/client/model.py
+++ b/src/euphorie/client/model.py
@@ -799,6 +799,27 @@ class SurveySession(BaseObject):
                     item.title = zodb_item.title
         self.refreshed = datetime.datetime.now()
 
+    def update_measure_types(self, survey):
+        """Update measure types in the session according to changes in the tool.
+        Specifically, if an `in_place_standard` measure is deleted in the tool, it
+        disappears from the identification phase of the session unless we change its
+        type to `in_place_custom`.
+        """
+        in_place_standard_measures = (
+            Session.query(Risk, ActionPlan)
+            .filter(Risk.id == ActionPlan.risk_id)
+            .filter(Risk.session_id == self.id)
+            .filter(ActionPlan.plan_type == "in_place_standard")
+            .all()
+        )
+        for risk, measure in in_place_standard_measures:
+            risk_zodb = survey.restrictedTraverse(risk.zodb_path.split("/"))
+            solution_ids_zodb = [sol.id for sol in risk_zodb._solutions]
+            if measure.solution_id not in solution_ids_zodb:
+                # The measure is in the session but not in the tool. It has probably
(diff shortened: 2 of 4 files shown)


src/euphorie/client/model.py
```diff
@@ -799,6 +799,20 @@ def refresh_survey(self, survey=None):
...
+            Session.query(Risk, ActionPlan)
+            .filter(Risk.id == ActionPlan.risk_id)
+            .filter(Risk.session_id == self.id)
+            .filter(ActionPlan.plan_type == "in_place_standard")
+            .all()
+        )
+        for risk, measure in in_place_standard_measures:
+            risk_zodb = survey.restrictedTraverse(risk.zodb_path.split("/"))
+            solution_ids = [sol.id for sol in risk_zodb._solutions]
+            if measure.solution_id not in solution_ids:
```
thet, line 813:
```suggestion
            if measure.solution_id not in solution_ids:
                # Measure not available because it was deleted in CMS.
                # Update type to ``in_place_custom`` do make them show up in existing sessions.
```

thet: I'd add a little more context information - but I'm not sure if my wording is correct...

src/euphorie/client/model.py
```diff
@@ -799,6 +799,20 @@ def refresh_survey(self, survey=None):
                     item.title = zodb_item.title
         self.refreshed = datetime.datetime.now()
 
+    def update_measure_types(self, survey):
```
thet, line 802:
```suggestion
    def update_measure_types(self, survey):
        """ Update measure type to ``in_place_custom`` if the type was
         ``in_place_standard`` but the measure deleted.
         This allows for showing deleted measures in old sessions where they were used.
         """
```

thet: More context information on what this method does. Not sure if the wording is correct.
The method name is a bit too generic for the specific purpose it serves, but feel free to keep it as it is.

src/euphorie/client/browser/session.py
```diff
@@ -329,8 +329,15 @@ def setupSession(self):
             # Touch means: the modification timestamp is set.
             # But we need to make sure the refreshed marker is up to date!
             survey_session.refresh_survey(survey)
-            return survey_session
+            new_session = survey_session
+        else:
+            new_session = self.rebuild_session(survey, profile)
 
+        new_session.update_measure_types(survey)
```
thet, line 336: I‌ looks a bit heavy to me to update all measure types for all risks every time one edits the Profile...
Would a upgrade step be better?
Not sure if this is worth the effort - it depends on how often the Profile edit view is used.

thet (changes requested):
One of the suggestions regarding comments in ``update_measure_types`` might be merged after the wording was corrected...

src/euphorie/client/browser/session.py
reinhardt: An upgrade step is meant to run once. We need code that runs every time a tool is changed and republished (specifically when an `in_place_standard` measure is deleted.

> I‌ looks a bit heavy to me to update all measure types for all risks every time one edits the Profile...

Please note that `profile` here refers to the profile questions on a tool. They determine which optional modules are active in the respective session. The profile view checks whether the structure of the tool has changed, because this might mean that the structure of the session needs to be adjusted.

We're not updating all measure types. The query specifically filters for `plan_type == in_place_standard`, and the type is only changed if a measure has been deleted.

I think this is pretty lightweight compared to the [`treeChanges`](https://github.com/euphorie/Euphorie/blob/97b14f3c41f3b9dc6b89e2641647115fa7500b24/src/euphorie/client/update.py#L90) method that runs every time the profile view is called, and especially the [rebuilding of the session](https://github.com/euphorie/Euphorie/blob/97b14f3c41f3b9dc6b89e2641647115fa7500b24/src/euphorie/client/browser/session.py#L334) that happens when `treeChanges` detects a change.

reinhardt:
I've added a docstring and comment and tweaked a variable name to hopefully make it clearer what's happening. What do you think?

src/euphorie/client/browser/session.py
thet: Thanks for the insight!

thet (approved):

Issue #379, opened 2022-03-31
The code that identifies the tab should be improved


ale-rt:
If the URL is not exactly what is expected the wrong tab will be highlited:
![image](https://user-images.githubusercontent.com/1300763/161029754-cf172b3c-bda2-4924-a58a-33ebb88bb1ba.png)

PR #433, opened 2022-08-11
Improved get_dashboard_tab

reinhardt:
Also made it possible to override the list of available tabs.

Refs syslabcom/scrum#497 euphorie/Euphorie#379


ale-rt:
IMO this is the wrong approach, we should do this using client side like discussed many times before.

See the way Plone does that: https://github.com/plone/mockup/blob/000884f123b3f0e09040960010be07ef8d21f657/src/pat/navigationmarker/navigationmarker.js

PR #437, opened 2022-09-09
Rich text (tiptap) for measures action

reinhardt:
Hide whitespace changes when checking the diff.

Refs syslabcom/scrum#444


ale-rt (commented):
Please check my comment.

Did you investigate the possible security implications of this change?

reinhardt:
> Did you investigate the possible security implications of this change?

No, thanks for pointing it out.

The only vector I can imagine is to add some malicious content like a javascript to the markup. That is currently stripped out before it is used in the training, but it should probably be stripped out on saving already. I noticed that the comment field (“Notes”) has the same issue.

reinhardt:
> That is currently stripped out before it is used in the training, but it should probably be stripped out on saving already.

This is now addressed in my latest commit.

> I noticed that the comment field (“Notes”) has the same issue.

I'll fix this in another PR.

reinhardt:
@ale-rt, do you have time to give this another quick look?

ale-rt (approved):

reinhardt:
:heart:

PR #442, opened 2022-09-15
Avoid upgrade error if training table already exists.

thet:


ale-rt (changes requested):
Please use the `has_table` helper:
https://github.com/euphorie/Euphorie/blob/a6e106588f174cf2ebe84ab7ca0e6e09eb1d99d6/src/euphorie/deployment/upgrade/alembic/versions/20220720110538_.py#L23

ale-rt (approved):

PR #444, opened 2022-09-15
pat-navigation updates

thet:
Use explicit pat-navigation trigger class on navigation elements which were pre Patternslib 9.2 also initialized.


ale-rt:
Does this mean that we can get rid of some Python code that set's the `current` CSS class?

ale-rt (approved):

thet:
Yes, that should now be possible

PR #445, opened 2022-09-16
Assessments filters

reinhardt:
Refs syslabcom/scrum#597

### DIFF
--- a/src/euphorie/client/browser/country.py
+++ b/src/euphorie/client/browser/country.py
@@ -383,16 +383,49 @@ class Assessments(BrowserView):
         return self.index
 
     @property
+    def organisation_options(self):
+        organisation_view = api.content.get_view(
+            "organisation", self.context, self.request
+        )
+        return [
+            {
+                "label": organisation.title,
+                "value": organisation.owner_id,
+                "selected": "selected"
+                if str(organisation.owner_id) == self.request.get("organisation", "")
+                else None,
+            }
+            for organisation in organisation_view.organisations
+        ]
+
+    def is_filter_active(self):
+        """True if any filters in the request parameters are different from the
+        defaults.
+        """
+        return (
+            self.request.get("organisation")
+            or self.request.get("sort_on", "alphabetical") != "alphabetical"
+        )
+
+    @property
     @memoize
     def sessions(self):
         searchable_text = self.request.get("SearchableText", None)
         if searchable_text and "%" not in searchable_text:
             searchable_text = "%{}%".format(searchable_text)
+        sort_on_value = self.request.get("sort_on", "alphabetical")
+        if sort_on_value == "alphabetical":
+            order_by = self.webhelpers.survey_session_model.title
+        else:
+            order_by = False
+        organisation_value = self.request.get("organisation", "")
         return self.webhelpers.get_sessions_query(
             context=self.context,
             searchable_text=searchable_text,
             include_archived=True,
-        ).all()
+            filter_by_account=organisation_value or True,
+            order_by=order_by,
+        )
 
 
 class Surveys(BrowserView, SurveyTemplatesMixin):
--- a/src/euphorie/client/browser/templates/assessments.pt
+++ b/src/euphorie/client/browser/templates/assessments.pt
@@ -15,7 +15,60 @@
     <metal:slot fill-slot="content">
       <div id="content-pane">
         <metal:homescreen_navigation use-macro="here/view/macros/homescreen_navigation">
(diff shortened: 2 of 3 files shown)


ale-rt:
There is a weird behavior.
If you set up some filters and go back to the search page you will see the filters active and the result filtered but the pannel will not show any filter:
image

I did not check if this is a design issue or some backend code thing

src/euphorie/client/browser/country.py
```diff
@@ -382,17 +384,83 @@ def form_action_name(self):
...
+            {
+                "label": account.title,
+                "value": account.id,
+                "selected": "selected"
+                if account.id == self.request.get("author", "")
+                else None,
+            }
+            for account, _ in accounts
+            if account is not self.webhelpers.get_current_account()
+        ]
```
ale-rt, line 408: I am afraid of the implications of this code against a production DB.
I fear the query might get a bit too expensive and slow.
We had something similar in the todo app and the list of accounts was replaced with an autoselect.
Anyway as you mention this is becoming a sort of alias of the organisation field.
I would ask @pilz to consider of leaving just the organisation field and to hide (at least for the moment this one).

src/euphorie/client/browser/templates/assessments.pt
```diff
@@ -15,7 +15,87 @@
...
       <div id="content-pane">
         <metal:homescreen_navigation use-macro="here/view/macros/homescreen_navigation">
-          <metal:slot fill-slot="view_options" />
+          <metal:slot fill-slot="view_options">
+            <div class="pat-collapsible pat-context-menu align-left no-label icon-filter closed"
+                 id="filter-menu"
+                 data-pat-collapsible="close-trigger: .context-menu:not(#filter-menu),.close-menu;"
+            >
+              <strong class="context-menu-label menu-trigger"
+                      id="filters-trigger"
```
ale-rt, line 24:
```suggestion
                      id="filters-trigger"
                      i18n:translate=""
```

ale-rt (changes requested):
Please initialize the value of the filters based on the request parameters

src/euphorie/client/browser/country.py
reinhardt: Alex said he could live without the author filter, so I'll take it out.

https://github.com/syslabcom/scrum/issues/597#issuecomment-1249183385

reinhardt: Come to think of it, there is currently no way that I can see a user's sessions if that user doesn't have an organisation (and has added me to that organisation), so `organisation` and `author` would be equivalent. It would only be beneficial to have the author name in case I know the person's name but not the organisation name, but I think this is highly unlikely.

One more reason to drop the `author` filter.

src/euphorie/client/browser/country.py
```diff
@@ -382,17 +382,50 @@ def form_action_name(self):
...
+        sort_on_value = self.request.get("sort_on", "alphabetical")
+        if sort_on_value == "alphabetical":
+            order_by = self.webhelpers.survey_session_model.title
+        else:
+            order_by = False
+        organisation_value = self.request.get("organisation", "")
         return self.webhelpers.get_sessions_query(
             context=self.context,
             searchable_text=searchable_text,
             include_archived=True,
```
ale-rt, line 425: Thinking out loud: having the include archived as a filter might be a nice enhancement. Anyway this is not in the scope of this task.
CC @pilz

src/euphorie/client/browser/country.py
```diff
@@ -382,17 +382,50 @@ def form_action_name(self):
...
+    def organisation_options(self):
+        organisation_view = api.content.get_view(
+            "organisation", self.context, self.request
+        )
+        return [
+            {
+                "label": organisation.title,
+                "value": organisation.owner_id,
+                "selected": "selected"
+                if organisation.owner_id == int(self.request.get("organisation", ""))
```
ale-rt, line 395: I would be more conservative, `int("")` will raise a `ValueError`, what about:
```suggestion
                if str(organisation.owner_id) == self.request.get("organisation", "")
```
it appears safer to me.

ale-rt (commented):
It looks well done, please check my concern about a possible raised `ValueError`

src/euphorie/client/browser/country.py
reinhardt: Good point! Let's do that.

ale-rt (approved):
LGTM, consider merging with squashed commits

PR #447, opened 2022-09-19
Dashboard tab marker is now set via pat-navigation.

thet:
Removed the backend code which sets the ``current`` class in the dashboard.
Instead it is set via ``pat-navigation`` client side only.

BEFORE‌ MERGING, please see:
https://gitlab.com/syslabcom/daimler/daimler.oira/-/merge_requests/144


ale-rt:
> Removed the backend code which sets the `current` class in the dashboard. Instead it is set via `pat-navigation` client side only.
> 
> BEFORE‌ MERGING, please see: https://gitlab.com/syslabcom/daimler/daimler.oira/-/merge_requests/144

From your analysis on that PR we have some cases that the previous code covered and this one did not

reinhardt:
On the one hand I think the behaviour with the dashboard switcher in daimler.oira is not ideal, on the other hand it seems to point to a general problem: In the backend we can easily distinguish between context and view; the javascript has to work on the URL where it isn't clear which names are context IDs and which are view names.

AFAICT merging this would not make the behaviour worse in plain Euphorie. It would make the code simpler (but only a little). We could still keep the old code for daimler.oira. However, that would introduce yet another deviation between the two.

If we don't merge this PR, will there be any conflict with the new pat-navigation javascript?

thet:
@ale-rt I‌ re-checked this PR with the current pat-navigation changes and it does work fine for me. I think we can merge this.

ale-rt:
I am not sure I am testing the right code, but before the patch after the login the dashboard was highlighted now it is not:
image
It seems to me that there is an issue handling the fragment.

BTW it is better if we have no tab highlighted rather than two.

PR #450, opened 2022-09-22
Fix navigation injection

reinhardt:
Make sure the #toolbar-functions-area is present because the nav injection needs it.

Best viewed [ignoring whitespace](https://github.com/euphorie/Euphorie/pull/450/files?diff=split&w=1)

Refs syslabcom/scrum#597


ale-rt (approved):
I approve this one but it feels bad

reinhardt:
I'll revisit this when Cornelis has fixed https://github.com/syslabcom/oira.prototype/issues/389

PR #451, opened 2022-09-22
pyupgrade pre commit hook

ale-rt:


ale-rt:
This is low priority but a nice thing to have

thet:
Did some changes: explicitly disabled the `code-analysis` pre commit hooks, moved the pyupgrade as first in the chain of pre-commits to allow black to format afterwards, did the reformatting of the whole code base in a separate commit.

thet (approved):
LGTM

PR #454, opened 2022-09-27
Upgrade to Patternslib 9.5.0.

thet:


reinhardt:
@thet, I don't see a good way to do a code review. Do you want me to do some manual testing before we merge instead?

thet:
Right, there is probably no point of reviewing this. We can also just merge.

reinhardt:
Yeah, let's!

PR #458, opened 2022-09-29
Improve the dashboard

ale-rt:


src/euphorie/client/browser/country.py
```diff
@@ -627,28 +627,35 @@ def sessions(self):
...
+            sessions = (
+                base_query.filter(account_id_column.notin_(known_owner_ids))
+                .limit(5)
+                .all()
+            )
+            if sessions:
+                sessions_by_organisation.insert(0, (None, sessions))
 
-        return sessions_by_organisation
+        return {key: value for key, value in sessions_by_organisation}
```
reinhardt, line 658: Just curious, doesn't `dict(sessions_by_organisation)` do the trick as well?

reinhardt (approved):

src/euphorie/client/browser/country.py
ale-rt: It does and it does it better:
```suggestion
        return dict(sessions_by_organisation)
```
Thanks!

PR #459, opened 2022-10-07
daimler fremdfirmen attachement

thet:
- theme update
- adding a generic ``#modal-container`` where modal dialoges can be injected.


ale-rt:
This one can be closed now, right?

PR #460, opened 2022-10-21
Added an hook to allow customizing who can see the public ``@@surveys`` view

ale-rt:


ale-rt:
I put the hold merge because I am evaluating if the view can be properly protected by a permission.

thet (approved):
LGTM

PR #461, opened 2022-10-21
Sanitize the permissions handling in the client section

ale-rt:
This [restores a line of code](https://github.com/euphorie/Euphorie/pull/461/files#diff-9687043f5fc4f5ff7e02e6778821127d070d35ee7d1c22e59d1cc50d155ebf4bL123) in the setup handler that publishes the client folder.
That line has been present for years and it is indeed needed for the application to work at all.
It was probably commented because the switch to the Plone 5 default test layer did not included setting up a default workflow chain.
This is now done in https://github.com/euphorie/Euphorie/pull/461/files#diff-972db14e3a8d2a845681d1db32e486a080300af841abac8d99d3d43181f3eb64R77-R78

With this change the test pass even if the "surveys" view is protected by the `zope2.View` permission, which is what we want to not leak unwanted informations.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -9,6 +9,8 @@ Changelog
 - Export/import training questions
   `#435 <https://github.com/euphorie/Euphorie/pull/435>`_.
 - Introduce the concept of "organisations"
+- Sanitize the permissions handling in the client section
+  [ale-rt]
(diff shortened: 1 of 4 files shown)


ale-rt:
This is a probe PR to see if the test pass with proper permission protection on the surveys view.
Refs. #460

reinhardt (approved):

PR #464, opened 2022-10-25
Investigate the surveys view permissions

ale-rt:


src/euphorie/client/model.py
```diff
@@ -566,9 +571,13 @@ def allowed(self, object, object_roles=None):
         if object_roles is None:
             return True
 
-        for role in ["Anonymous", "Authenticated", "EuphorieUser"]:
-            if role in object_roles:
-                return True
+        if not isinstance(object_roles, (list, tuple)):
```
ale-rt, line 574: To my understanding `object_roles` should always be an iterable.
Is this check you added just needed to support the existing tests?
If yes please revert and modify the tests.

src/euphorie/client/tests/test_model.py
```diff
@@ -17,6 +21,11 @@
     import mock
 
 
+@implementer(IClient)
+class MockContext(Acquisition.Implicit):
```
ale-rt, line 25: There should be no need to have a mocked context here.
The portal should have an existing clients folder

reinhardt (approved):
I can't judge whether it does what's needed, but the code looks clean and simple enough to me.

ale-rt:
With @pilz we agreed that this will be merged after the next release

src/euphorie/client/model.py
```diff
@@ -566,9 +571,10 @@ def allowed(self, object, object_roles=None):
         if object_roles is None:
             return True
 
-        for role in ["Anonymous", "Authenticated", "EuphorieUser"]:
-            if role in object_roles:
-                return True
+        for obj in aq_chain(aq_inner(context)):
+            if IClient.providedBy(obj):
+                allowed_roles = {"Anonymous", "Authenticated", "EuphorieUser", "Reader"}
```
thet, line 576: Shouldn't this be `allowed_roles = self.getRoles()`?
If we do that, we can also customize `getRoles` for different deployments.
Not sure, if I got this right, though.

ale-rt: They are different things.
`getRoles` is there because this class act as a role PAS plugin.
`allowed`  is there because this class act as a PAS member object.

ale-rt: As you can see the tests show this is working as expected

PR #467, opened 2022-10-26
Update resources

thet:
Update oira.prototype JS resources only.

Upgrade to Patternslib 9.7.0-alpha.5 which fixes wrong display of pat-markdown content, which is now fixed:
https://oira.cornelis.amsterdam/ass-private-security-grafton-street/items/2014-01-03-slips-trips-and-falls-are-prevented


ale-rt (commented):
I think it is fine, ask @pilz for a green light before merging

PR #471, opened 2022-11-02
Upgrade to Plone 5.2.10

ale-rt:


ale-rt:
@pilz I am unsure if we want to merge this now.
All depends on when the DTMB release will happen.

PR #482, opened 2022-11-23
Rich text for omega risk measures

reinhardt:
Use rich text (tiptap) for measures in place on omega risks as well.

Separate label for text area to avoid checkbox stealing focus.
Wrapping `<template>` to avoid duplicated element.

Refs syslabcom/scrum#753


ale-rt (approved):
I approve but I have no clue :)

PR #488, opened 2022-12-20
Lazily load the dashboard content

reinhardt:
This feels a little silly - maybe we should lazy-load the individual portlets instead. But this does work around the certificate size issue (syslabcom/scrum#712).

See https://github.com/syslabcom/oira.prototype/commit/2c236988ec5ef44cd6a5ec9e86855eed83f9f6ac

syslabcom/scrum#712

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -16,7 +16,7 @@ Changelog
   [ale-rt]
 - Decoupled loading of the dashboard.
   Refs: `#712 <https://github.com/syslabcom/scrum/issues/712>`_.
-  [reinhardt]
+  [reinhardt, thet]
 
 
 14.2.1 (2022-11-25)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -14,6 +14,9 @@ Changelog
   [reinhardt]
 - Upgrade plone to 5.2.10
   [ale-rt]
+- Decoupled loading of the dashboard.
+  Refs: `#712 <https://github.com/syslabcom/scrum/issues/712>`_.
+  [reinhardt]
 
 
 14.2.1 (2022-11-25)
--- a/src/euphorie/client/browser/configure.zcml
+++ b/src/euphorie/client/browser/configure.zcml
@@ -412,6 +412,15 @@
       />
 
   <browser:page
+      name="dashboard"
+      for="euphorie.client.country.IClientCountry"
+      permission="zope2.View"
+      class=".country.SessionsView"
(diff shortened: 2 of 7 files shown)


pilz:
We do lazy load the portlets on quaive already. This can be a long term goal. For this issue, this is enough.

reinhardt:
I've pushed a fix for one of the test failures, but most seem unrelated to my changes:

`NotImplementedError: (<Control name='tiptap-href' type='url'>, <Field name="tiptap-href">)`

pilz:
Pull hannes in

thet:
@reinhardt please check my fixes.

Please note: any request parameters would get lost when calling the `@@dashboard` view.  It looks like, there are some which might be used: https://github.com/euphorie/Euphorie/blob/main/src/euphorie/client/browser/country.py#L273
We could solve that by adding a method to the `SessionsView` class which creates the URL for the `@@dashboard` view including any relevant request parameters.

reinhardt:
Well spotted, thanks! And there may be more for daimler.

thet:
@reinhardt I put that on draft until to prevent a merge before we have resolved the mentioned issues.

reinhardt:
> It looks like, there are some which might be used: https://github.com/euphorie/Euphorie/blob/main/src/euphorie/client/browser/country.py#L273

The `action` parameter is actually only used by the view `new-session.html` which registers the same class `SessionsView`.

Instead we need to take care of `hide_archived` and `hide_archived_marker` which are used by `portlet-my-ras` (`MyRAsPortlet`): https://github.com/euphorie/Euphorie/blob/dashboard-autoload/src/euphorie/client/browser/country.py#L594

Daimler has multiple variations of this portlet plus a search box in the "Our Contractors" portlet.

reinhardt:
I'd like to go a step or two further and lazy-load the individual portlets instead of a view that renders the whole dashboard content area. That way we can also point any forms in portlets (which set the parameters like `hide_archived`) to views that only render a single portlet, which gives us another performance boost.

Skipping the intermediate step of making the `@@dashboard` view aware of all the request parameters that the portlets may need could save us some effort.

This is also in line with our efforts toward componentizing quaive proto at SlcCon in December.

reinhardt:
I've whipped up a proposal for lazy-loading individual portlets in proto: https://github.com/syslabcom/oira.prototype/pull/402

reinhardt:
Replaced by https://github.com/euphorie/Euphorie/pull/494

PR #490, opened 2023-01-02
Add the @@country-tools view to the site menu.

thet:
Together:
https://github.com/euphorie/NuPlone/pull/42
https://github.com/euphorie/Euphorie/pull/490
https://github.com/euphorie/osha.oira/pull/180

Fixes: https://github.com/syslabcom/scrum/issues/837

### DIFF
--- a/src/euphorie/deployment/browser/sitemenu.py
+++ b/src/euphorie/deployment/browser/sitemenu.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 from AccessControl import getSecurityManager
+from AccessControl import Unauthorized
 from Acquisition import aq_inner
 from euphorie.content import MessageFactory as _
 from euphorie.content.sector import ISector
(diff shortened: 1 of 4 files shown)


src/euphorie/deployment/browser/sitemenu.py
```diff
@@ -10,9 +10,47 @@
...
+            return menu
+        else:
+            return None
+
+    def menu_country_tools(self):
+        context = aq_inner(self.context)
+        request = self.request
+
+        # We try to render the view. It would fail for the wrong context (not
+        # an ICountry) or if permissions are not met.
```
reinhardt, line 38: I'm pretty sure `queryMultiAdapter` doesn't check permissions. In this case it doesn't matter much because `country-tools` is registered with `zope2.View`. But I would suggest either not mentioning permissions in the comment or double checking that permissions are actually respected (e.g. by writing a test for that).

reinhardt (changes requested):

src/euphorie/deployment/browser/sitemenu.py
thet: That was basically copying the logic from https://github.com/euphorie/osha.oira/blob/7645c3d5dbe8fc157b02f1c042c5d91bf7b738d8/src/osha/oira/content/sitemenu.py#L31
But I‌ think you're right. I'm checking that.

reinhardt: While we're at it, querying the adapter doesn't render the view. Maybe rephrasing that would be good as well.

... and the comment in osha.oira should probably be changed as well. Wow, that's 10 years old...

thet: done.

reinhardt (approved):

PR #491, opened 2023-01-03
CSV download of country tools

reinhardt:
syslabcom/scrum#838

### DIFF
--- a/src/euphorie/content/browser/country.py
+++ b/src/euphorie/content/browser/country.py
@@ -4,11 +4,15 @@ from euphorie.content.countrymanager import ICountryManager
 from euphorie.content.sector import getSurveys
 from euphorie.content.sector import ISector
 from euphorie.content.utils import CUSTOM_COUNTRY_NAMES
+from io import StringIO
 from plone.dexterity.browser.add import DefaultAddForm
 from plone.dexterity.browser.add import DefaultAddView
 from plone.dexterity.browser.edit import DefaultEditForm
 from plone.memoize.instance import memoize
 from Products.Five import BrowserView
+
+import csv
+import datetime
 
 
 class CountryView(BrowserView):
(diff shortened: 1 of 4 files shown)


src/euphorie/content/browser/country.py
```diff
@@ -101,3 +104,38 @@ class Tools(CountryView):
...
+        fieldnames = [
+            "Sector",
+            "Tool Title",
+            "Version Title",
+            "URL",
+            "published",
+            "obsolete",
+        ]
+        buffer = StringIO()
+        writer = csv.DictWriter(buffer, fieldnames=fieldnames, dialect="bilbomatica")
```
thet, line 118: why the dialect `bilbomatica`?‌ the default is `excel` which would be what users most often expect and what the ticket description asks for.
If there is a reason, please leave a comment in code.

src/euphorie/content/browser/country.py
```diff
@@ -101,3 +104,38 @@ class Tools(CountryView):
...
+                            "published": "published" if survey["published"] else "",
+                            "obsolete": "obsolete" if group["obsolete"] else "",
+                        }
+                    )
+        csv_data = buffer.getvalue()
+        buffer.close()
+        response = self.request.RESPONSE
+        response.setHeader(
+            "Content-Disposition",
+            f"attachment; filename=oira_tools_{self.context.id}.csv",
```
thet, line 138: just a matter of taste, should we include the date isostring also in the filename? I personally would do so.

src/euphorie/content/browser/templates/country_tools.pt
```diff
@@ -13,6 +13,7 @@
     <metal:title fill-slot="title">${view/title}</metal:title>
     <metal:content fill-slot="content">
       <h2 i18n:translate="header_sector_survey_list">OiRA Tools</h2>
+      <a class="button" href="${context/absolute_url}/@@country-tools.csv" i18n:translate="header_sector_survey_download">Download spreadsheet</a>
```
thet, line 16: :+1: :)

thet (changes requested):
All fine, except for clarification for the csv dialect and a suggestion which you may just ignore.

src/euphorie/content/browser/country.py
reinhardt: I went for consistency with [existing code](https://github.com/euphorie/Euphorie/blob/country-tools-csv/src/euphorie/content/utils.py#L27), but thinking about it again, you're probably right and the default is better if nothing specific was requested.

reinhardt:
Pushed new version with default dialect and date in file name.

thet (approved):
:+1:

PR #492, opened 2023-01-03
Organisation: Add user text tweak

reinhardt:
When adding a user, don't mention trainings if they are disabled.

syslabcom/scrum#841

### DIFF
--- a/src/euphorie/client/browser/templates/panel-add-user-to-organisation.pt
+++ b/src/euphorie/client/browser/templates/panel-add-user-to-organisation.pt
@@ -49,7 +49,7 @@
                            type="radio"
                            value="member"
                     />
-                    <tal:i18n i18n:translate="">Member</tal:i18n>
+                    <tal:i18n i18n:translate="label_role_member">Member</tal:i18n>
                   </label>
                   <label tal:condition="nothing">
                     # XXX commented for the moment because it is still under discussion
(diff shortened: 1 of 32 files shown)


src/euphorie/client/browser/templates/panel-add-user-to-organisation.pt
```diff
@@ -66,24 +66,47 @@
...
                   </label>
-                  <p class="pat-message info pat-depends"
-                     data-pat-depends="role=admin"
-                     i18n:translate=""
-                  >
+
+                  <tal:training_enabled tal:condition="view/is_training_enabled">
+                    <p class="pat-message info pat-depends"
+                       data-pat-depends="role=admin"
+                       i18n:translate=""
```
thet, line 73: I do prefer explicit msgids. That keeps already translated strings still valid even if a little text is changed in the original string.

thet (approved):
LGTM.

As noted in the comment, I would use explicit msgid like `i18n:translate="label_admin_role__training_enabled"` which makes the tranlated strings robust against any changes in the original msgstr.
But that would mean to manually add these msgid's to the current translations and also, this is to some extend a bit of a question of taste.

Approved.

src/euphorie/client/browser/templates/panel-add-user-to-organisation.pt
reinhardt: In general I agree. I judged it not worth the effort to change the existing strings in this case.

reinhardt: Then again, I might as well do that now while there's not much else going on.

reinhardt: Two things:

* i18ndude doesn't seem to support changing msgids. It marks the old translation as fuzzy and adds a new empty one. Fixing that manually for a dozen languages is tedious. Am I doing that wrong?
* It's good that I looked into it because it made me notice that the panel for editing organisation members uses the same texts and should be kept in sync.

reinhardt:
I won't change the msgids that are already translated for now (see above). Do you think using symbolic msgids for the new texts makes things more confusing, or is that fine? @thet

thet:
@reinhardt I think adding symbolic msgids for new texts is totally fine!

PR #493, opened 2023-01-04
Update CI infrastructure.

thet:
This PR updates versions, cleans up buildout, adds NuPlone to versions.cfg and uses the NuPlone release instead of the source checkout.
That should improve the extensibility of the versions.cfg in oira.application.buildout, osha.oira or other projects.

Note: this removes the `versions-plone-5.2-latest.cfg` file. I have no idea why this is included here. It's tedious to keep up-to-date and since we need network access to install Plone anyways, I don't see a point of maintaining it.
Maybe I'm missing something?


reinhardt:
> Note: this removes the `versions-plone-5.2-latest.cfg` file. I have no idea why this is included here. It's tedious to keep up-to-date and since we need network access to install Plone anyways, I don't see a point of maintaining it.
> Maybe I'm missing something?

I'd like to ask @ale-rt about that. He often knows that kind of stuff.

reinhardt:
I found this:

> We actually do need local copies of the plone and zope versions since they are using http to reference each other

https://github.com/euphorie/Euphorie/commit/7f2872409e2d100f654ce97d4afcc4278d51aee2

reinhardt (changes requested):
When running buildout with `devel.cfg` (e.g. via the Makefile) I get the error

```
While:
  Installing sphinx.
Error: The requirement ('Sphinx>=4.0') is not allowed by your [versions] constraint (1.8.6)
```

reinhardt:
> When running buildout with `devel.cfg` (e.g. via the Makefile) I get the error
> 
> ```
> While:
>   Installing sphinx.
> Error: The requirement ('Sphinx>=4.0') is not allowed by your [versions] constraint (1.8.6)
> ```

It can be fixed by pinning `repoze.sphinx.autointerface = 0.8`, which this PR removes from `versions.cfg`. You may want to review the changes to that file.

thet:
> I found this:
> 
> > We actually do need local copies of the plone and zope versions since they are using http to reference each other
> 
> [7f28724](https://github.com/euphorie/Euphorie/commit/7f2872409e2d100f654ce97d4afcc4278d51aee2)

Thanks for pointing to that. This is not the case anymore, the version files are referenced via https only.

thet:
@reinhardt added the repoze.sphinx.autointerface version fixture.

reinhardt (approved):

PR #494, opened 2023-01-05
Lazy-load the dashboard portlets

reinhardt:
See also https://github.com/syslabcom/oira.prototype/pull/402

Replaces #488

In #488 @thet noticed this:

> Please note: any request parameters would get lost when calling the @@dashboard view.

I found that request parameters were only set by `portlet-my-ras` which contains a form for toggling the display of archived items. It was re-injecting the whole dashboard. Now it only injects itself; therefore the dashboard does not need to be aware of request parameters.

For variants with custom dashboards like daimler we need to double check that the portlets also only re-inject themselves.

syslabcom/scrum#712

### DIFF
--- a/src/euphorie/client/browser/templates/sessions-dashboard.pt
+++ b/src/euphorie/client/browser/templates/sessions-dashboard.pt
@@ -107,13 +107,27 @@
              data-pat-masonry="column-width: .grid-sizer; gutter: 0; item-selector: .portlet; stamp: .stamp;"
         >
           <div class="grid-sizer"></div>
-          <tal:portlets repeat="portlet view/portlets"
-                        replace="structure portlet"
-          />
+          <tal:portlets repeat="portlet view/portlets">
+            <div class="portlet span-${portlet/columns|string:1}"
+                 id="${portlet/element_id}"
+                 tal:condition="portlet/element_id|nothing"
+            >
+              <a class="pat-inject infinite-scrolling-trigger"
+                 href="${context/absolute_url}/@@${portlet/__name__}#${portlet/element_id}"
+                 data-pat-inject="trigger: autoload"
+              >Loading&hellip;</a>
+            </div>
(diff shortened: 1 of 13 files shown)


src/euphorie/client/browser/templates/sessions-dashboard.pt
```diff
@@ -107,13 +107,27 @@
...
-                        replace="structure portlet"
-          />
+          <tal:portlets repeat="portlet view/portlets">
+            <div class="portlet span-${portlet/columns|string:1}"
+                 id="${portlet/element_id}"
+                 tal:condition="portlet/element_id|nothing"
+            >
+              <a class="pat-inject infinite-scrolling-trigger"
+                 href="${context/absolute_url}/@@${portlet/__name__}#${portlet/element_id}"
+                 data-pat-inject="trigger: autoload"
```
thet, line 117: I wonder if `trigger: autoload-visible` should be used here. This adds a small JavaScript overhead though.

src/euphorie/client/browser/templates/sessions-dashboard.pt
```diff
@@ -107,13 +107,27 @@
...
+              >Loading&hellip;</a>
+            </div>
+            <tal:comment replace="nothing"><!-- Fallback: If the view has no element_id member, we can't properly pre-render the div --></tal:comment>
+            <a class="pat-inject infinite-scrolling-trigger"
+               href="${context/absolute_url}/@@${portlet/__name__}"
+               data-pat-inject="trigger: autoload; target: self::element"
```
thet, line 123: Same here.

thet (approved):
LGTM. Two remarks about possibly using `trigger: autoload-visible`, but only if it makes sense, e.g. if we can expect the portlets to be below the fold.

reinhardt:
I'd say the idea of a dashboard is to have an immediate overview of important information, so I would load everything right away even before any interaction like scrolling.

That said, we should keep it aligned with proto; let's see how https://github.com/syslabcom/scrum/issues/712 pans out. However, I will already merge this one because it makes so much sense. We can still adapt it later.

PR #495, opened 2023-02-15
Fix: Logo and Hamburger overlap mobile + markup hour

thet:
Ref: scrum-953

In surveys in mobile view, there is now a tree toggler:

image

Expanded view:

image

This PR‌ contains this change plus some simplifications and bug fixes.

Please see the individual commits.


ale-rt (approved):
LGTM

src/euphorie/client/browser/webhelpers.py
```diff
@@ -1170,5 +1170,146 @@ def is_new_session(self):
...
+                "href": f"{url}/@@start#content",
+                "title": api.portal.translate(
+                    _("label_preparation", default="Preparation")
+                ),
+                "has_tree": False,
+            }
+        )
+
+        if self.use_involve_phase:
+            active = section == "involve"
```
reinhardt, line 1205: Reusing the `active` and `disabled` variables might lead to bugs when setting them is omitted by accident.

reinhardt (approved):
Just one small thing that I'll leave to your discretion.

src/euphorie/client/browser/webhelpers.py
thet: tnx for the review, but I‌ think I'll keep it as it is currently.
Let's do it in another step.
I'm happy for simplifications of this method anyways!

PR #501, opened 2023-03-02
training view: re-add measures key in slide_contents.

thet:
The change in https://github.com/euphorie/Euphorie/pull/500 was breaking subclasses of the training view.


src/euphorie/client/browser/training.py
```diff
@@ -233,12 +233,13 @@ def image_urls(self):
                 )
         return urls
 
-    def slide_contents(self, standalone=False):
```
thet, line 236: that parameter is nowhere used. Therefore i removed it in a separate commit.

reinhardt (approved):

PR #503, opened 2023-03-03
HtmlToWord: Fix problem with different documents for each report compilation

thet:
This involves an API change: Instantiate HtmlToWord compiler freshly for each document and make use of oject orientated encapsulation. Also do not call the compiler but create a dedicated compile_inline method.

This fixes a problem where a bullet number counter was not reset among multiple docx compilations, resulting in different documents for the same report when downloaded multiple times.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -3,6 +3,13 @@ Changelog
 
 14.3.1 (unreleased)
 -------------------
+
+- HtmlToWord: Fix problem with different documents for each report compilation.
+  This fixes a problem where a bullet number counter was not reset among
+  multiple docx compilations, resulting in different documents for the same
+  report when downloaded multiple times.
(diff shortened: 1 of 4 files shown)


docs/changes.rst
```diff
@@ -3,6 +3,16 @@ Changelog
 
 14.3.1 (unreleased)
 -------------------
+
+- HtmlToWord: Fix problem with different documents for each report compilation.
+  This involves an API change: Instantiate HtmlToWord compiler freshly for each
```
reinhardt, line 8: If we want this to be the customer-facing change log then I would leave out the technical information.

reinhardt (approved):

docs/changes.rst
thet: :+1:

PR #509, opened 2023-03-08
Demo with the whole form autosubmit

ale-rt:


ale-rt:
For the moment this cannot be merged because the newly injected measures are not included in the form.

PR #518, opened 2023-03-22
Restrict the assessments view to authenticated users

ale-rt:
Refs. https://github.com/syslabcom/scrum/issues/1043


thet (approved):
LGTM

I also tested registering a new test user and still could access the assessments page. Fine!

PR #523, opened 2023-04-05
Redirect to new session URL when tool is updated

reinhardt:
I'm not sure whether this is the right approach because it requires extensive changes due to the fact that euphorie disables the redirection storage...

syslabcom/scrum#833


ale-rt:
What a surprise we found in the Easter egg :) It would be nice to know why the redirection tool was disabled.
Maybe @wichert or @pysailor remember that.

In any case I am prone to think about an alternative solution.

reinhardt:
Replaced by https://github.com/euphorie/Euphorie/pull/524

PR #524, opened 2023-04-06
Keep session URL usable when tool is updated

reinhardt:
This currently keeps the URL with the old session ID. I'd rather do a redirect to the URL with the new ID but I can't figure out how to do that in a namespace traversal. `response.redirect()` just gives me a blank page, `raise Redirect` gives me a traceback. Any ideas?

syslabcom/scrum#833

### DIFF
--- a/src/euphorie/client/model.py
+++ b/src/euphorie/client/model.py
@@ -1211,8 +1211,7 @@ class SessionRedirect(BaseObject):
 
     __tablename__ = "session_redirect"
 
-    id = schema.Column(types.Integer(), primary_key=True, autoincrement=True)
-    old_session_id = schema.Column(types.Integer(), unique=True, nullable=False)
+    old_session_id = schema.Column(types.Integer(), primary_key=True, nullable=False)
     new_session_id = schema.Column(types.Integer(), nullable=False)
 
 
--- a/src/euphorie/client/model.py
+++ b/src/euphorie/client/model.py
@@ -1206,6 +1206,16 @@ class SurveySession(BaseObject):
         return completion_percentage
 
 
+class SessionRedirect(BaseObject):
+    """Mapping of old deleted sessions to their new rebuilt counterparts"""
+
+    __tablename__ = "session_redirect"
+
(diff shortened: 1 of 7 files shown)


src/euphorie/client/model.py
```diff
@@ -1206,6 +1206,16 @@ def completion_percentage(self):
         return completion_percentage
 
 
+class SessionRedirect(BaseObject):
+    """Mapping of old deleted sessions to their new rebuilt counterparts"""
+
+    __tablename__ = "session_redirect"
+
+    id = schema.Column(types.Integer(), primary_key=True, autoincrement=True)
```
ale-rt, line 1214: Do we need the `id` here?
I would say that the primary key is the `old_session_id`

ale-rt:
What if you handle the `Redirect` exception in the `error_view`?
Or maybe there is something like an `IBeforeTraversal` adapter that can be used for that.
Anyway if it takes too much effort finding the proper solution go with an MVP.
It is already a massive improvement and so far no customer complained.
And the URLs in the page should have the new id.
It does not seem a big deal to me anyway.

src/euphorie/client/model.py
reinhardt: When I tried to make the `old_session_id` the primary key (without `autoincrement`) I got the error

```
sqlalchemy.exc.ArgumentError: Mapper mapped class SessionRedirect->session_redirect could not assemble any primary key columns for mapped table 'session_redirect'
```

so I added a separate primary key.

reinhardt:
Thanks! I'll invest an hour more into exploring these hints.

src/euphorie/client/model.py
reinhardt: I can look into it again if you think it's worth it.

reinhardt:
I've found a way to do it via an error view.

src/euphorie/client/model.py
reinhardt: Actually it works with `old_session_id`. Not sure what caused the error previously, but it must have been something else.

ale-rt (approved):
LGTM, but I did not test it TTW :)

PR #531, opened 2023-04-18
Fixed French translation of “edit”

reinhardt:
My French is not excellent but I'm pretty sure “Créer” is “Create” and not “Edit”. Plone core also has “Modifier” for “Edit”.

syslabcom/scrum#594


reinhardt:
Lia says via email to change it and see whether someone complains.

PR #536, opened 2023-04-21
Request validation from consultant

reinhardt:
Includes #533; merge that one first

https://github.com/syslabcom/scrum/issues/1105

### DIFF
--- a/src/euphorie/client/model.py
+++ b/src/euphorie/client/model.py
@@ -27,6 +27,7 @@ from sqlalchemy import func
 from sqlalchemy import orm
 from sqlalchemy import schema
 from sqlalchemy import sql
+from sqlalchemy import Table
 from sqlalchemy import types
 from sqlalchemy.event import listen
 from sqlalchemy.ext import declarative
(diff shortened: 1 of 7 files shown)


reinhardt:
Hold on, I think I need to do some more tweaks...

src/euphorie/client/model.py
```diff
@@ -651,6 +652,22 @@ class ISurveySession(Interface):
     """Marker interface for a SurveySession object."""
 
 
+consultancy = Table(
```
ale-rt, line 655: Why do you declare the Table in this fancy format?

reinhardt: To avoid modifiying the SurveySession and Account tables themselves. If we need to remove the consultancy stuff for some reason then we can simply drop this table and don't have to fiddle with the original ones.

It follows this pattern: https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html#many-to-many

... which is under “Basic Relationship Patterns”, so it's not fancy but basic. :wink:

ale-rt (approved):
LGTM

PR #540, opened 2023-04-26
Deactivate organisation invite confirmation mail

reinhardt:
Follow-up of #539 

Deployment postponed because translations are missing.

syslabcom/scrum#1133


ale-rt (approved):
It looks not really ok to merge something like this, but ok

PR #542, opened 2023-04-26
Fix the translation in the title attribute of listed assessments

ale-rt:
Fixes this weird msgid: https://github.com/euphorie/Euphorie/blob/2cdd6a50d2e6966ef631c1ec0115323f149760dc/src/euphorie/deployment/locales/euphorie.pot#L49


ale-rt:
Replaced by #543

PR #545, opened 2023-04-26
Fix some bits in the email top the consultant

ale-rt:


ale-rt:
@reinhardt this PR fixes the mail text that is sent out in current master.
image
As you can see, the HTML email is sent as text.
The purpose is to convert back the entities to characters.

Also it removes the `meta:interpolation` which is not needed.

reinhardt:
It's not an HTML mail though, but `text/plain`. #538 changed the quotes to be utf-8 characters instead of HTML entities. So if we want an HTML mail we need to change that back and also change the mime type, right?

2023-04-27-13-45-17_1837x1570

ale-rt:
Never mind

PR #546, opened 2023-05-02
Validation: Record status of validation

reinhardt:
Original plan was to treat a locked session with a consultant assigned as validated, but we need to support locking sessions independently of validation status.

This also means that the sqlalchemy declaration via `Table` doesn't work any more because the association needs the extra column `status`.

https://github.com/syslabcom/scrum/issues/1105

### DIFF
--- a/src/euphorie/client/browser/configure.zcml
+++ b/src/euphorie/client/browser/configure.zcml
@@ -893,7 +893,7 @@
       name="consultancy"
       for="euphorie.client.adapters.session_traversal.ITraversedSurveySession"
       permission="euphorie.client.ViewSurvey"
-      class=".consultancy.Consultancy"
+      class=".consultancy.ConsultancyView"
       template="templates/consultancy.pt"
       layer="euphorie.client.interfaces.IClientSkinLayer"
       />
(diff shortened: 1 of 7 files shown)


ale-rt:
If you want, merge that, I am not really following.
But it seems to me we are moving to a wrong direction.
IMO, it is better to have the `SessionEvent` table and record the status there.
We can discuss this, maybe also involving @pilz

reinhardt:
I'm not fond of the SessionEvent table for storing status information. To efficiently get a list of sessions containing information on when the last briefing was sent and whether sampling is due requires all kinds of weird joins (see [`get_sessions_query`](https://gitlab.com/syslabcom/daimler/daimler.oira/-/blob/master/src/daimler/oira/client/browser/country.py#L773)).

That will improve if we introduce an additional state variable (like the Plone workflow also has).

I'm available for further discussion, of course.

reinhardt:
> IMO, it is better to have the `SessionEvent` table and record the status there.
> 
> The idea behind that table is to mimic the workflow history we have on regular content objects.
> 
> We can discuss this, maybe also involving @pilz

For discussion: https://github.com/euphorie/Euphorie/pull/553

PR #553, opened 2023-05-04
Validation: Record session events

reinhardt:
To do: Backport the session history from daimler.oira and integrate SessionValidated and SessionValidationRequested.

syslabcom/scrum#1162

### DIFF
--- a/src/euphorie/client/model.py
+++ b/src/euphorie/client/model.py
@@ -431,8 +431,8 @@ class Consultancy(BaseObject):
         primary_key=True,
     )
     account_id = schema.Column(
-        schema.ForeignKey("account.id", onupdate="CASCADE", ondelete="CASCADE"),
-        primary_key=True,
+        schema.ForeignKey("account.id", onupdate="CASCADE", ondelete="SET NULL"),
+        nullable=True,
     )
 
     session = orm.relationship(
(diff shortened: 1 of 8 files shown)


src/euphorie/client/model.py
```diff
@@ -1248,6 +1248,41 @@ def completion_percentage(self):
...
+        "SurveySession",
+        cascade="all,delete-orphan",
+        single_parent=True,
+        backref=orm.backref("event", uselist=False, cascade="all"),
+    )
+    action = schema.Column(types.Unicode(32))
+    note = schema.Column(types.Unicode)
+    file_data = schema.Column(types.LargeBinary())
+    uuid_responsible = schema.Column(types.String(32), nullable=True)
+    uuid_alternate = schema.Column(types.String(32), nullable=True)
```
ale-rt, line 1283:
```suggestion
```
The table can be extended in other projects

src/euphorie/client/tests/test_consultancy.py
```diff
@@ -110,6 +113,21 @@ def test_request_validation(self):
...
                 )
+                events = (
+                    self.session.query(SessionEvent)
+                    .filter(
+                        SessionEvent.session_id == self.traversed_session.session.id
+                    )
+                    .order_by(SessionEvent.time.desc())
+                    .all()
+                )
+                self.assertEqual(len(events), 1)
```
ale-rt, line 124: If you take the query above and call `.one()` instead of `.all()`, you are automatically asserting that you have one matching row.

reinhardt: That would change the test failure to a database error. I don't think that's what I want.

ale-rt: That's the point, you will not even assert... anyway, that's fine

src/euphorie/client/tests/test_consultancy.py
```diff
@@ -202,24 +225,83 @@ def test_validate_risk_assessment(self):
...
+            mail_fixture.storage[1][0][1],
+        }
+        self.assertSetEqual(
+            recipients,
+            {"valerie@labyrinth.social", "jessica@labyrinth.social"},
+        )
+
+    def test_delete_consultant(self):
+        now = datetime.utcnow().replace(tzinfo=timezone.utc)
+        self.traversed_session.session.consultancy = Consultancy(
```
ale-rt, line 269: It seems to me we should not have this column but just rely on the validation event.

ale-rt (approved):
I am quite out of the topic, you are much more in to that. Feel free to merge

ale-rt (approved):
I am quite out of the topic, you are much more in to that. Feel free to merge

src/euphorie/client/tests/test_consultancy.py
reinhardt: I'm not sure which column you mean but see https://github.com/euphorie/Euphorie/pull/546#issuecomment-1532642945 for why I don't want to rely on the event alone.

ale-rt:
@reinhardt please check if the tests here are broken because of this PR being merged https://ci.syslab.com/job/daimler/1394/console

ale-rt:
I am taking care of that

PR #557, opened 2023-05-04
Optional training module

ale-rt:
When there is no measure to select, do not show the "Training" module.

contains #556


ale-rt:
RA **with no** additional measures:
image
RA **with** additional measures:
image

reinhardt (approved):

PR #558, opened 2023-05-04
Fix some bits in the email top the consultant

ale-rt:
I am resurrecting #545 because I still see that the mail sent looks like:
```
Dear b@example.com,
  Ale requested to validate the risk assessment &lsquo;COVID-19 revision&rsquo;.

  Please click on the link below to visit the risk assessment. If you approve of the contents of the risk assessment, then please click on &lsquo;Validate&rsquo; in the section &lsquo;Consultancy&rsquo; to give the risk assessment an official validated status.

  http://localhost:9110/Plone/client/eu/covid-19/covid-19-revision/++session++2/@@start
```

That happens because zpretty escaped the quote characters.
Either we make zpretty ignore that file or we change zpretty to not do that substitution or we do it like proposed in this PR.

### DIFF
--- a/src/euphorie/client/browser/consultancy.py
+++ b/src/euphorie/client/browser/consultancy.py
@@ -9,6 +9,7 @@ from plone import api
 from plone.memoize.view import memoize
 from Products.Five.browser.pagetemplatefile import ViewPageTemplateFile
 
+import html
 import logging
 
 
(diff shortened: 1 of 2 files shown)


reinhardt (approved):
It's a bit weird to use an HTML function on non-html text, but it looks like it'll work and is quick and easy.

PR #566, opened 2023-05-22
Don't show consultant role if consultancy phase is disabled

reinhardt:
syslabcom/scrum#1225

### DIFF
--- a/src/euphorie/client/browser/organisation.py
+++ b/src/euphorie/client/browser/organisation.py
@@ -38,25 +38,33 @@ class OrganisationBaseView(BaseView):
 
     default_target_view = "@@organisation"
 
-    _known_roles = [
-        {
-            "value": "member",
-            "label": _("Member"),
-        },
-        # XXX commented for the moment because it is still under discussion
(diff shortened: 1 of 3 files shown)


src/euphorie/client/browser/organisation.py
```diff
@@ -192,6 +200,13 @@ def publishTraverse(self, request, organisation_id):
         request.set(self.organisation_id_key, organisation_id)
         return self
 
+    @property
+    def role_options(self):
+        """Return a list of options for the role field."""
+        options = deepcopy(self._known_roles)
+        options[0]["checked"] = "checked"
+        return options
```
ale-rt, line 208: Is it needed to have this property, AFAICS you can just add `checked: "checked"` in line 44

ale-rt (approved):

src/euphorie/client/browser/organisation.py
reinhardt: No, that will mess up editing existing members (line 746).

PR #570, opened 2023-05-24
Consider a session locked if the consultant has validated it

ale-rt:


reinhardt:
refs syslabcom/scrum#1161

reinhardt (approved):
Will have to be changed for syslabcom/scrum#1097 but is simple enough that it's fine to merge now

PR #575, opened 2023-05-29
Rename the published feature to locking and make it more powerful

ale-rt:
Ref.: scrum-1097


src/euphorie/client/browser/webhelpers.py
```diff
@@ -1004,6 +1028,50 @@ def is_owner(self):
         session = self.traversed_session.session
         return self.get_current_account() == session.account
 
+    @property
+    @memoize
+    def is_manager(self):
+        """Check if the current user is the owner of the session."""
```
reinhardt, line 1034:
```suggestion
        """Check if the current user is a manager of the session."""
```

src/euphorie/client/model.py
```diff
@@ -801,20 +802,42 @@ def is_archived(self):
             return False
         return archived <= localized_now()
 
+    @property
+    def last_locking_event(self):
+        """Return the last event relative to locking"""
+        query = (
+            Session.query(SessionEvent)
+            .filter(
+                SessionEvent.action.in_(("lock_set", "lock_reset", "lock_unset")),
```
reinhardt, line 811: Use `LOCKING_ACTIONS` from config

```suggestion
                SessionEvent.action.in_(LOCKING_ACTIONS),
```

src/euphorie/client/model.py
```diff
@@ -801,20 +802,42 @@ def is_archived(self):
...
         consultancy = self.consultancy
-        if not consultancy:
+        if consultancy and consultancy.status == "validated":
+            return True
+
+        event = self.last_locking_event
+        if not event:
             return False
-        return consultancy.status == "validated"
+        return event.action in ("lock_set", "lock_reset")
```
reinhardt, line 828:
```suggestion
        return event.action in LOCKING_SET_ACTIONS
```

src/euphorie/client/model.py
```diff
@@ -801,20 +802,42 @@ def is_archived(self):
...
+        )
+        return query.first()
+
     @property
     def is_locked(self):
+        """Check if the session is locked."""
         consultancy = self.consultancy
-        if not consultancy:
+        if consultancy and consultancy.status == "validated":
+            return True
```
reinhardt, line 823:
```suggestion
```

I'd suggest explicitly setting the lock in the consultancy view so that you don't need another case distinction here.

Alternatively add the validated event to the list of locking events (but that seems like a violation of separation of concerns).

src/euphorie/client/browser/locking.py
```diff
@@ -0,0 +1,77 @@
+from euphorie.client import model
+from plone import api
+from plone.memoize.view import memoize
+from plone.memoize.view import memoize_contextless
+from Products.Five import BrowserView
+from zExceptions import Unauthorized
+
+
+class LockingMenu(BrowserView):
+    lock_actions = ["lock_set", "lock_reset", "lock_unset"]
```
reinhardt, line 10:
```suggestion
    lock_actions = LOCKING_ACTIONS
```

src/euphorie/client/config.py
```diff
@@ -1,3 +1,6 @@
 FULL_ACCOUNT = "full"
 GUEST_ACCOUNT = "guest"
 CONVERTED_ACCOUNT = "converted"
+
+LOCKING_SET_ACTIONS = ("lock_set", "lock_reset")
```
reinhardt, line 5: Why `lock_reset`? “Reset” could be interpreted as the opposite of “set”. Additionally it does not match the method name `refresh_lock`.

I'd rather use `lock_refresh` or `lock_refreshed`.

src/euphorie/client/browser/session.py
```diff
@@ -792,34 +793,39 @@ def __call__(self):
...
 
+    @deprecate("Use the @@locking_view instead. Deprecated in version 15.0.0.dev0")
     def reset_date(self):
         """Reset the session date to now."""
-        session = self.context.session
-        session.published = datetime.now()
-        session.last_publisher = get_current_account()
-        return self.redirect()
+        target = f"{self.context.absolute_url()}/@@locking_view/refresh_lock"
+        return self.redirect(target)
```
reinhardt, line 804: It looks like you're deprecating the feature and changing it at the same time. Isn't deprecation meant to give someone who uses the feature time to adapt before it is changed or removed?

src/euphorie/client/browser/webhelpers.py
```diff
@@ -1004,6 +1028,50 @@ def is_owner(self):
...
+    @property
+    @memoize
+    def can_unlock_session(self):
+        if not self.use_locking_feature:
+            return False
+        if self.is_owner:
+            return True
+        if self.is_manager:
+            return True
+        return False
```
reinhardt, line 1073:
```suggestion
        return self.can_lock_session
```

reinhardt (changes requested):

src/euphorie/client/model.py
ale-rt: This has indeed to change, but not by setting the lock but by checking if a "validate sesssion event" has been recorded.
Don't you agree?

src/euphorie/client/browser/session.py
ale-rt: I added the deprecation to the UI, allowing to still use it with the updated backend.
There is an upgrade step that updates the data in such a way that this deprecated UI is still functional but behaves like an alias for the new logic

src/euphorie/client/model.py
reinhardt: As I said, that seems like a violation of separation of concerns. The locking code should not have to check the validation status of a session. Also, in the long run an admin should be able to unlock a session even if it is validated.

src/euphorie/client/browser/session.py
reinhardt: Not sure about it but let's leave it like this.

src/euphorie/client/model.py
reinhardt: We can tweak this later though.

reinhardt (approved):

PR #577, opened 2023-05-30
Improvements on the locking story

ale-rt:
- Do not use the consultancy status to check if a session is locked
- Only show the actions in the locking menu if the session is not validated and if the user has the proper permissions

### DIFF
--- a/src/euphorie/client/browser/locking.py
+++ b/src/euphorie/client/browser/locking.py
@@ -34,6 +34,20 @@ class LockingMenu(BrowserView):
         """Return whether the session is locked."""
         return self.context.session.is_locked
 
+    def is_validated(self):
+        """Return whether the session is locked."""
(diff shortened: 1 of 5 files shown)


src/euphorie/client/browser/locking.py
```diff
@@ -34,6 +34,20 @@ def is_locked(self):
         """Return whether the session is locked."""
         return self.context.session.is_locked
 
+    def is_validated(self):
+        """Return whether the session is locked."""
```
thet, line 38:
```suggestion
        """Return whether the session is validated."""
```
... or is locked equivalent to validated?

thet (approved):
LGTM. Only one comment was confusing to me, when in the method `is_validated` the comment says `is locked`.
Please consider fixing that, otherwise LGTM

PR #578, opened 2023-05-31
Keep consultancy phase enabled when assessment is locked

reinhardt:
syslabcom/scrum#1161

### DIFF
--- a/src/euphorie/client/browser/webhelpers.py
+++ b/src/euphorie/client/browser/webhelpers.py
@@ -1358,6 +1358,8 @@ class WebHelpers(BrowserView):
                 }
             )
 
+        # The following menu items should be active even if can_edit is False.
+        disabled = section in ("", "preparation") and is_new_session
         if self.use_consultancy_phase:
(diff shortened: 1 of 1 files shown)


src/euphorie/client/browser/webhelpers.py
```diff
@@ -1377,8 +1379,6 @@ def survey_tree_data(self):
 
         # TODO: check if guest.
         active = section == "report"
-        # The following menu items should be active even if can_edit is False.
-        disabled = section in ("", "preparation") and is_new_session
```
ale-rt, line 1381: Looking at the code this was always `False` so removing these two lines is anyway good right?

ale-rt (approved):
Looks sane

src/euphorie/client/browser/webhelpers.py
reinhardt: They are not removed but moved up by a few lines. The `disabled` variable is reused for multiple sections (which, granted, is not ideal in terms of readability).

PR #579, opened 2023-05-31
Allow unlocking validated assessments

reinhardt:
syslabcom/scrum#1161

### DIFF
--- a/src/euphorie/client/browser/consultancy.py
+++ b/src/euphorie/client/browser/consultancy.py
@@ -256,6 +256,13 @@ class PanelValidateRiskAssessment(ConsultancyBaseView):
             )
             self.sqlsession.add(event)
             self.context.session.consultancy.status = "validated"
+            if not self.context.session.is_locked:
+                locking_view = api.content.get_view(
+                    name="locking_view",
+                    context=self.context,
+                    request=self.request,
+                )
+                locking_view.create_event("lock_set")
             self.notify_admins()
         self.redirect()
 
--- a/src/euphorie/client/browser/webhelpers.py
+++ b/src/euphorie/client/browser/webhelpers.py
@@ -1064,9 +1064,6 @@ class WebHelpers(BrowserView):
     @property
     @memoize
     def can_unlock_session(self):
-        session = self.traversed_session.session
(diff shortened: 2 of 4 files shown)


src/euphorie/client/browser/consultancy.py
```diff
@@ -256,7 +256,12 @@ def handle_POST(self):
             )
             self.sqlsession.add(event)
             self.context.session.consultancy.status = "validated"
-            # TODO: lock session
+            locking_view = api.content.get_view(
+                name="locking_view",
+                context=self.context,
+                request=self.request,
+            )
+            locking_view.create_event("lock_set")
```
ale-rt, line 264: I do not think we need to create a lock event, we already have the validated event

reinhardt: The validated event is not in the list of locking events, so it is ignored by the locking view.

ale-rt: You probably missed https://github.com/euphorie/Euphorie/pull/577

reinhardt: Actually I think we want a condition here. If the assessment has already been locked by an admin before it is validated, then we don't want to lock it again here. We want to show the time and user name from the original locking event in the locking menu.

reinhardt: > You probably missed #577

I am aware of it but it does not fulfill the requirements.

ale-rt: Can you please update your PR to cope with the fact that a lock can be there either because there was a "lock_set" action and a "validated" action?

reinhardt: I've updated the code. Please see the tests for why we need it this way and ping me if you need further explanation.

reinhardt:
Under discussion - do you lose the validation when unlocking?

pilz:
Validation is like signature and stamp on an official document. What has been validated must not be changed anymore.

reinhardt:
syslabcom/scrum#1137 says otherwise, please discuss there

reinhardt:
Replaced by #612

PR #586, opened 2023-06-01
Populate link “Find an OSH consultant”

reinhardt:
syslabcom/scrum#1164

### DIFF
--- a/src/euphorie/client/browser/configure.zcml
+++ b/src/euphorie/client/browser/configure.zcml
@@ -974,6 +974,15 @@
       layer="euphorie.client.interfaces.IClientSkinLayer"
       />
 
+  <browser:page
+      name="consultants"
+      for="euphorie.client.adapters.session_traversal.ITraversedSurveySession"
+      class=".consultancy.Consultants"
+      template="templates/consultants.pt"
+      permission="zope2.View"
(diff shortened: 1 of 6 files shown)


ale-rt:
What is content expected to be?
I suppose a `euphorie.page`.

If yes please check https://github.com/euphorie/Euphorie/pull/587. 

It was necessary to me toactually display this page without having a:
```
  Module Products.CMFDynamicViewFTI.browserdefault, line 88, in __call__
RecursionError: maximum recursion depth exceeded in comparison
```

ale-rt (approved):
Ok.
Is there a reason why you are not using portal catalog to find objects with `getId="consultants"` ?

reinhardt:
I know exactly where the objects are, so there is no need to find them.

Any reason why I should use the portal catalog?

ale-rt:
> I know exactly where the objects are, so there is no need to find them.
> 
> Any reason why I should use the portal catalog?

Because you are iterating over `objectValues` and that can be potentially lots of objects to wake up before you get the proper one. 
In addition, there are security quirks.

That should be just a minor problem in a real world scenario, in fact I approved the PR.

Just consider these remarks for the next time.

reinhardt:
Great, thanks for elaborating.

PR #592, opened 2023-06-02
Fix selected measure text spanning whole page instead of bein shown in training module.

thet:
Ref: scrum-1205 item 2)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,6 +4,10 @@ Changelog
 15.0.0 (unreleased)
 -------------------
 
+- Fix selected measure text spanning whole page instead of bein shown in training module.
+  Ref: scrum-1205 item 2)
+  [thet]
+
 - Align with prototype to fix form style issues.
   Ref: scrum-1205
   [thet]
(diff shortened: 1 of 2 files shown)


reinhardt:
This seems a bit hacky. Can't we make the injection target depend on whether the training module is displayed?

thet:
Actually, I'd do it is shown in my commit, because it works better than expected:

The injection target replaces the whole element: `source: #page-module-training::element; target: #page-module-training::element;`. When first rendered with no risks selected it is hidden. When a risk is selected and the injection happens the result already does not contain the `display:hidden` style and the risk module is shown immediately. When unselecting the risk module is hidden immediately.

If we would fiddle with the injection depending on if there are selected risks or not we would only show/hide the risk module on a full reload/injection of the page.

Slightly related: what I found odd is:

1) that when selecting pat-autosubmit POSTs the change to the server and the server responds already with the correct markup. But then another injection happens just to GET the same markup and to inject/replace the risk module.

2) besides of being loaded twice, the whole page is returned, not only relevant parts of it. We could save quite some templating logic when only the relevant parts would be returned. Or - instead of the `style="display: none"` hack if we could use the CSS `:has` pseudo selector instead‌ (not yet available in Firefox) and instead of pat-inject only pat-clone the selected item into the training module.

pilz:
Yes but that would complicate the code considerably. There is no funding for making this cleaner. Keep as is.

PR #593, opened 2023-06-02
Duplication: Tools Section

reinhardt:
Also use updated modal when starting a new session from the tools section

Note: there is one quirk - when calling the new-session modal on a sector (which does not happen in the current code) and passing the tool (survey) as a request parameter, you can select all sessions from the sector, not only those from the tool. Not fixing because the current code does not do that and sectors mostly only have one tool.

syslabcom/scrum#1137

### DIFF
--- a/src/euphorie/client/browser/configure.zcml
+++ b/src/euphorie/client/browser/configure.zcml
@@ -615,6 +615,14 @@
       />
 
   <browser:page
+      name="assessments.json"
+      for="euphorie.content.survey.ISurvey"
+      class=".country.AssessmentsJson"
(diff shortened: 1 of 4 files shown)


thet (commented):
Code wise this does look good to me. Your comment about selecting any session from the sector sounds a bit scary. Could that be a security or data protection problem?

reinhardt:
> Could that be a security or data protection problem?

No. Permissions are handled per individual session, so you always only see your sessions. The filtering by sector would be merely for convenience.

Sorry I didn't make that clearer.

PR #596, opened 2023-06-07
Show solution descriptions in the Word report

reinhardt:
except for France and Italy when measures in place are active

syslabcom/scrum#1245

### DIFF
--- a/src/euphorie/client/docx/compiler.py
+++ b/src/euphorie/client/docx/compiler.py
@@ -167,6 +167,8 @@ class BaseOfficeCompiler:
 class DocxCompiler(BaseOfficeCompiler):
     _base_filename = "oira.docx"
 
+    use_solution_description = True
+
     @property
     def _template_filename(self):
         return resource_filename(
@@ -194,6 +196,11 @@ class DocxCompiler(BaseOfficeCompiler):
         self.tool_type = get_tool_type(self.webhelpers._survey)
         self.tti = getUtility(IToolTypesInfo)
         self.italy_special = self.webhelpers.country == "it"
+
+    @property
+    @memoize
+    def survey(self):
+        return self.context.aq_parent
 
     def set_session_title_row(self, data):
         """This fills the workspace activity run with some text."""
@@ -447,7 +454,20 @@ class DocxCompiler(BaseOfficeCompiler):
             ]
 
         m = measure
-        values = [_simple_breaks(m.action or "")]
+        action = m.action
+        if self.use_solution_description and measure.plan_type in [
+            "in_place_standard",
+            "measure_standard",
+        ]:
+            action = "\n".join(
+                (
+                    self.survey.restrictedTraverse(
+                        "/".join((measure.risk.zodb_path, measure.solution_id))
+                    ).description,
+                    action,
+                )
+            )
+        values = [_simple_breaks(action or "")]
         if not implemented:
             values = values + [
                 _simple_breaks(m.requirements or ""),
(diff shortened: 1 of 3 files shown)


src/euphorie/client/docx/compiler.py
```diff
@@ -447,7 +454,20 @@ def add_measure(self, doc, heading, measure, implemented=False):
             ]
 
         m = measure
-        values = [_simple_breaks(m.action or "")]
+        action = m.action
```
ale-rt, line 457: To reduce the lines of code of this method and to save an `if` in an already complex method, what about having something like:
```
action = self.measure2action(measure)
```

On the base class `measure2action` returns `measure.action` but for the others it returns the code in the if block.

I hope you get the point.

src/euphorie/client/docx/views.py
```diff
@@ -111,23 +111,32 @@ def get_risks_for(self, sql_module):
...
             risk_description = _sanitize_html(risk_description)
 
-            measures = [
-                measure.action
-                for measure in (
-                    sql_risk.in_place_standard_measures
-                    + sql_risk.in_place_custom_measures
-                )
-            ]
+            # XXX can we move this distinction to the compiler?
```
ale-rt, line 114: That seems a good idea

src/euphorie/client/docx/views.py
```diff
@@ -111,23 +111,32 @@ def get_risks_for(self, sql_module):
...
-                    _get_action_plan(action)
-                    for action in (
-                        sql_risk.standard_measures + sql_risk.custom_measures
+                    _get_action_plan(
+                        action, solution=risk.get(action.solution_id, None)
                     )
+                    for action in (sql_risk.standard_measures)
                 ]
+                actions.extend(
+                    [_get_action_plan(action) for action in (sql_risk.custom_measures)]
```
ale-rt, line 138:
```suggestion
                    [_get_action_plan(action) for action in sql_risk.custom_measures]
```

src/euphorie/client/docx/views.py
```diff
@@ -111,23 +111,32 @@ def get_risks_for(self, sql_module):
...
                 risk and getattr(risk, "type", None) == "top5"
             ):
                 actions = [
-                    _get_action_plan(action)
-                    for action in (
-                        sql_risk.standard_measures + sql_risk.custom_measures
+                    _get_action_plan(
+                        action, solution=risk.get(action.solution_id, None)
                     )
+                    for action in (sql_risk.standard_measures)
```
ale-rt, line 135:
```suggestion
                    for action in sql_risk.standard_measures
```

ale-rt (approved):
LGTM and I am ok to merge this one.

It would be cool if you could check my suggestions.

Generally speaking, I think it is a good practice to not introduce feature flags and if all over the place to handle custom behaviors because it often leads to code which is harder to maintain.

If you decide not to rework this PR, please bear in mind for the next time.

reinhardt:
Thanks for the suggestions. I see even more potential to improve this code. I'll check how much effort we can spend.

reinhardt:
> Generally speaking, I think it is a good practice to not introduce feature flags and if all over the place to handle custom behaviors because it often leads to code which is harder to maintain.

What would you say is a better alternative?

ale-rt:
> > Generally speaking, I think it is a good practice to not introduce feature flags and if all over the place to handle custom behaviors because it often leads to code which is harder to maintain.
> 
> What would you say is a better alternative?

See my suggestion in https://github.com/euphorie/Euphorie/pull/596#discussion_r1221339895

You code looks like

```python

class A:
    flag_active = False

    def __call__(self):
        if self.flag_active:
            return "something"
        else:
            return "something else"

class B(A):
    flag_active = True
```

My suggestion is to make it look like to this:

```python

class A:
    def __call__(self):
        return "something"

class B(A):
    def __call__(self):
        return "something else"
```

That's especially helpful in this case because of the cyclomatic complexity of the method.

reinhardt:
Oh, that's what you mean. Now I get it, thanks. I do see the benefits. On the other hand, if I have a flag with a speaking name like `use_solution_descriptions`,  it is potentially clearer what the intention is.

reinhardt:
I'll merge this now and clean up in a new pull request.

PR #597, opened 2023-06-09
SQLAlchemy 1.4

ale-rt:


src/euphorie/client/model.py
```diff
@@ -31,9 +31,8 @@
 from sqlalchemy import sql
 from sqlalchemy import types
 from sqlalchemy.event import listen
-from sqlalchemy.ext import declarative
-from sqlalchemy.orm import backref
-from sqlalchemy.orm import relationship
+from sqlalchemy.ext.declarative.extensions import instrument_declarative
+from sqlalchemy.orm.decl_base import _declarative_constructor
```
reinhardt, line 35: Is it really necessary to import a private object? This might be unstable, right?

src/euphorie/client/model.py
```diff
@@ -686,6 +685,14 @@ class AccountChangeRequest(BaseObject):
     )
 
 
+Account.change_request = orm.relationship(
```
reinhardt, line 688: Why are you declaring this (and some other attributes)  outside of the class?

src/euphorie/client/tests/test_model.py
```diff
@@ -495,7 +495,7 @@ def test_get_account_filter(self):
         )
         self.assertEqual(
             str(session.get_account_filter([account, "2"])),
-            "session.account_id IN (:account_id_1, :account_id_2)",
+            "session.account_id IN (__[POSTCOMPILE_account_id_1])",
```
reinhardt, line 498: This looks scary. What is that `POSTCOMPILE` prefix and what does the `__`  do?

src/euphorie/client/model.py
ale-rt: Was it more stable before?
Do you require to change it?

src/euphorie/client/model.py
ale-rt: When referencing a column, the relation requires to be defined after the referenced class is defined.

src/euphorie/client/tests/test_model.py
ale-rt: They are just strings like they were before.
How can the new code be scarier than the old code?

src/euphorie/client/model.py
reinhardt: Or you can reference it by name (string).

reinhardt:
If you know what you're doing, go ahead. I'm not fully following.

src/euphorie/client/model.py
ale-rt: I know that trick of using strings for the table class.

I tried:
- `account_id`
- `SurveySession.account_id`
- `session.account_id`

sqlalchemy.exc.ArgumentError: Column expression expected for argument 'foreign_keys'; got '$THE_STRING_I_TRIED'.

I did not dig deeper in this 
Feel free to make another PR to fix that if you know how to do it.

BTW, the SQLAlchemy documentation also has a similar example: https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html#adding-relationships-to-mapped-classes-after-declaration

It suggests:
```
Parent.children = relationship(Child, primaryjoin=Child.parent_id == Parent.id)
```
which is more or less the same thing I am doing...

PR #603, opened 2023-06-14
Plone6

ale-rt:


.github/workflows/main.yml
```diff
@@ -39,15 +41,17 @@ jobs:
...
           sudo locale-gen en_US.UTF-8 nl_NL@euro
           python -m venv .
+          ln -s requirements-${{ matrix.config[2] }}.txt requirements.txt
           bin/pip install -r requirements.txt
+          ln -s versions-${{ matrix.config[2] }}.cfg versions.cfg
           bin/buildout -c .github.cfg
       - name: Pre commit
         uses: pre-commit/action@v2.0.0
-        if: ${{ matrix.config[1] == '3.8' }}
+        if: ${{ matrix.config[1] == '3.11' }}
```
thet, line 50: 3.11 is not in the matrix.

.github/workflows/main.yml
```diff
@@ -39,15 +41,17 @@ jobs:
...
       - name: Code Analysis
         run: |
           bin/code-analysis
-        if: ${{ matrix.config[1] == '3.8' }}
+        if: ${{ matrix.config[1] == '3.11' }}
```
thet, line 54: same, 3.11 is not in the matrix.

src/euphorie/content/risk.py
```diff
@@ -660,7 +659,6 @@ def fixed_priority(self, value):
         self.default_priority = value
 
     @property
-    @memoize
```
thet, line 663: I'm fine with using memoize sparsely and keep this change, but what problem does this solve?
Seems unrelated to the Plone 6 changes here.

thet (changes requested):

.github/workflows/main.yml
ale-rt: The 3.11 is planned but commented some lines above

.github/workflows/main.yml
ale-rt: The 3.11 is planned but commented some lines above

thet (approved):

PR #604, opened 2023-06-15
Exclude the short report from the country exceptions

reinhardt:
syslabcom/scrum#1245


src/euphorie/client/docx/views.py
```diff
@@ -308,7 +308,10 @@ def _filename(self):
 
 
 class ActionPlanShortDocxView(ActionPlanDocxView):
-    _compiler = DocxCompilerShort
+    @property
+    @memoize
+    def _compiler(self):
+        return DocxCompilerShort
```
ale-rt, line 314: This customization should not be needed, anyway good job.

Please double check if this can removed. In that case clean up and merge :)

ale-rt (approved):

src/euphorie/client/docx/views.py
reinhardt: In the `ActionPlanShortDocxView` I always want to use the `DocxCompilerShort` which is never used in `ActionPlanDocxView`. This will not be the case if I remove this override. Or am I misunderstanding something?

ale-rt, line 314: What I mean is that leaving the code as it was, i.e.:
```suggestion
    _compiler = DocxCompilerShort
```
has to work as well.


```
>>> class A:
...     @property
...     def foo(self): return 1
... 
>>> class B:
...     foo = 2
... 
>>> A().foo
1
>>> B().foo
2
```

ale-rt (approved):

PR #612, opened 2023-06-19
Allow invalidating a validated assessment

reinhardt:
syslabcom/scrum#1302


reinhardt:
Probably requires more fixes

src/euphorie/client/browser/templates/locking_badge.pt
```diff
@@ -4,23 +4,45 @@
...
+    <tal:case case="default">
+      <a class="icon-lock-open icon pat-tooltip"
+         href="${here/absolute_url}/@@locking_menu#locking-menu"
+         title="Lock this assessment"
+         data-pat-tooltip="source: ajax; position-list: tr; trigger: click"
+         i18n:translate="label_unlocked"
+      >Unlocked
+      </a>
+    </tal:case>
+  </tal:switch>
```
ale-rt, line 47: Am I wrong, or the session can be both locked and validated?
I would keep this possibility at least.

reinhardt: If it is validated, then it is always also locked. You cannot unlock it without invalidating it, and when you invalidate it, it is also unlocked. So the three possible states are unlocked, locked, and validated. From the validated state you can only go to the unlocked state. From the other states you can go anywhere.

ale-rt: > If it is validated, then it is always also locked

I tend to disagree here... 
I also discussed that with @pilz and IIRC we agreed the two things are orthogonal.

When it is validated, it is not locked. It is just not editable.

Should we talk about that again?

ale-rt: Also remember that we already have customers that use the locking, and for them nothing should change.

pilz: unlocked -> locked -> validated (also not editable)
similar to private -> submitted -> published. 
There is no need to go from validated to unlocked. I find it more suitable to go from validated to locked because the whole path is an edge case and if it happens, I would like it to be as much controlled as possible. And therefore as explicit as possible. Each step back should be a conscious step. The owner of the organisation should invalidate (Use case that just someone else should validate again). And if the owner decides that he want to also change something, he should explicitly unlock (use case that a change is necessary).

Let me know if that helps.

reinhardt: @pilz, will you take this to Cornelis to update the design?

reinhardt: I've updated this PR so that after invalidating an assessment it is still locked, as discussed this morning.

ale-rt (approved):

PR #618, opened 2023-07-03
Remove some text from the help section

thet:
This PR removes some text from the help section as requested here:
https://github.com/syslabcom/scrum/issues/1267#issuecomment-1617734445

It is based on a branch of oira.prototype:
https://github.com/syslabcom/oira.prototype/tree/scrum-1267--help--remove-some-text

This needs to be deployed, but I'm unsure if we should merge these branches as cornelis is continuing to work on prototype master and we need to have these text removed until the user management feature is ready.

@angeldasangel what's your opinion on merging/keeping these branches?


reinhardt:
I would appreciate keeping a fixed branch from which we pull prototype updates (Is master the best choice for that? Discuss!) because otherwise it's easy to make a mistake next time someone updates resources from prototype.

thet:
I think this is obsolete since: https://github.com/syslabcom/oira.prototype/commit/2fd7de2187a504ba790618c5cb34df5190ffa038

PR #620, opened 2023-07-05
Change the name of the action

ale-rt:
This is consistent to what is done in the lock menu: https://github.com/euphorie/Euphorie/blob/0e6635b7d33eba30e05558a715bf25da9929cb5d/src/euphorie/client/config.py#L5
https://github.com/euphorie/Euphorie/blob/0e6635b7d33eba30e05558a715bf25da9929cb5d/src/euphorie/client/browser/locking.py#L72


reinhardt (commented):
LGTM

PR #623, opened 2023-07-10
Run make pot

ale-rt:
I ran just `make pot` in the root of the package Euphorie, I think I never did that.

I am not sure why we have so many strange changes, e.g.:
```diff
-"Language: deployment\n"
+"Language: bg\n"
```

```diff
-#: Euphorie/src/euphorie/client/browser/templates/login.pt:327
+#: euphorie/client/browser/templates/login.pt:327
```

Maybe I am not doing that properly?

PR #624, opened 2023-07-10
Makefile: No dependency for target `pot`

reinhardt:
I'd like to be able to run `make pot` without running buildout first. I can take care of installing i18ndude myself.


reinhardt:
See also #623

ale-rt (approved):

PR #625, opened 2023-07-12
Similar titles

ale-rt:


src/euphorie/content/browser/similar_titles.py
```diff
@@ -0,0 +1,181 @@
...
+        description=_("Maximum similarity between titles"),
+        default=1.0,
+        required=True,
+        min=0.0,
+        max=1.0,
+    )
+
+
+class SimilarTitles(AutoExtensibleForm, form.Form):
+    """The upload view for a :obj:`euphorie.content.sector`
```
reinhardt, line 44: Docstring does not seem correct - copy & paste?

src/euphorie/content/browser/similar_titles.py
```diff
@@ -0,0 +1,181 @@
...
+    def min_similarity(self):
+        return self.extracted_data.get("min_similarity", 0.5)
+
+    @property
+    def max_similarity(self):
+        return self.extracted_data.get("max_similarity", 1.0)
+
+    @property
+    def tool_cache(self):
+        return {}
```
reinhardt, line 87: I don't think I get it... Will this not return a new dict on every call? Why make it an instance method then?

src/euphorie/content/browser/similar_titles_details.py
```diff
@@ -0,0 +1,38 @@
+from euphorie.content import MessageFactory as _
+from euphorie.content.surveygroup import ISurveyGroup
+from plone import api
+from Products.Five import BrowserView
+
+
+class SimilarTitlesDetails(BrowserView):
+    """The upload view for a :obj:`euphorie.content.sector"""
```
reinhardt, line 8: Docstring does not seem correct - copy & paste?

src/euphorie/content/browser/similar_titles_details.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @property
+    def tool_cache(self):
+        return {}
```
reinhardt, line 12: See above

src/euphorie/content/browser/similar_titles.py
ale-rt: I think this misses a `memoize`. Isn't this already a

reinhardt: Yeah, that was my question: Why did you make it an instance method? I.e. why not explicitly use `{}` in the method below.

... but now it makes sense! :+1:

reinhardt (approved):

PR #646, opened 2023-09-29
Missing translations MOI-670

angeldasangel:


reinhardt:
There were some clashes with translations added as part of #1574. I gave the translations in the current PR precedence.

reinhardt (approved):

PR #649, opened 2023-10-10
More short report tweaks.

reinhardt:
Keep top/bottom cell borders between risks of same module.

syslabcom/scrum#1295

### DIFF
--- a/src/euphorie/client/docx/compiler.py
+++ b/src/euphorie/client/docx/compiler.py
@@ -1082,8 +1082,8 @@ class DocxCompilerShort(DocxCompilerFullTable):
 
     def merge_module_rows(self, row_module, row_risk):
         """Override to not remove the horizontal borders between the unmerged cells."""
-        for idx in range(1):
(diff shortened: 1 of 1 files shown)


ale-rt (changes requested):
`for idx in range(1) ` looks like code obfuscation to me.

Is that a leftover?

ale-rt (commented):
Apparently it comes from this one https://github.com/syslabcom/daimler.oira/commit/b7d0d4642586c434adc437cc0fbae19d7f230096#diff-611f13a21a8050fbb7ed577516423f0b6fb62298c842d004a2b6c7252ce232f2R159-R161, which at least explains the reason of the strange `range(1)`.
It would be nice to have clearer code/docs.
The future developers will really be pleased.

reinhardt:
I got rid of the loop.

reinhardt:
Here's another PR to also change the original bit of code: #650

ale-rt (approved):

PR #655, opened 2023-10-19
Make alembic upgrades fail-safe.

thet:
In order to run my own alembic upgrade I had to make all other not to fail first :unamused: 

With these changes here I could finally run my upgrade steps.

Co-authored by GitHub Copilot, that saved me at least from some typing.


ale-rt:
You should use alembic stamp to avoid running uneeded upgrade step

thet:
Which I did. For some reason it seems to to have worked...

reinhardt (approved):

PR #659, opened 2023-10-25
Add View find-solution-title-duplications

reinhardt:
syslabcom/scrum#1550

### DIFF
--- a/src/euphorie/content/browser/survey.py
+++ b/src/euphorie/content/browser/survey.py
@@ -448,3 +448,18 @@ class ContentsOfSurvey(IdentificationReportDocxView):
             mapping=dict(title=self.context.title),
         )
         return f"{translate(filename, context=self.request)}.docx"
+
+
+class FindSolutionTitleDuplications(BrowserView):
+    """"""
+
+    @property
+    def measures(self):
+        brains = api.content.find(
+            portal_type="euphorie.solution",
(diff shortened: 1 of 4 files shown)


src/euphorie/content/browser/survey.py
```diff
@@ -448,3 +448,18 @@ def _filename(self):
...
+
+    @property
+    def measures(self):
+        brains = api.content.find(
+            portal_type="euphorie.solution",
+            path="/".join(self.context.getPhysicalPath()),
+        )
+        for brain in brains:
+            obj = brain.getObject()
+            if obj.description in obj.action:
```
thet, line 464: shouldn't it be the other way around? assuming that `obj.action` is the title...

```suggestion
            if obj.action in obj.description:
```

thet (changes requested):
I think I have found something, but I'm not sure. If the code is OK, feel free to merge.

Note: this will not do the check if the the title in the beginning of the text would be necessary for the semantic of the description. But this wasn't asked for in the first step either.

src/euphorie/content/browser/survey.py
reinhardt: No, this is an oira weirdness. What the ticket calls the title is in the `description` attribute and what it calls the description is actually the `action` attribute.

PR #660, opened 2023-10-26
Add view find-tools-with-duplications

reinhardt:
syslabcom/scrum#1550


thet (approved):
Approved. One minor change suggestion, though.

PR #676, opened 2023-12-01
Notifications: Don't fail on data inconsistency when retrieving the s…

thet:
…ession url.

### DIFF
--- a/src/euphorie/client/notifications/base.py
+++ b/src/euphorie/client/notifications/base.py
@@ -48,6 +48,10 @@ class BaseNotificationEmail(BaseEmail):
         """The mail text."""
         full_name = self.webhelpers.get_user_fullname(self.account)
 
+        title_missing = api.portal.translate(
+            _("label_missing_title", default="Title is missing")
+        )
+
         session_links = ""
         for session in self.sessions:
             session_url = ""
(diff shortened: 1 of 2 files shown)


src/euphorie/client/notifications/base.py
```diff
@@ -43,12 +47,24 @@ def recipient(self):
...
+            session_url = ""
+            try:
+                session_url = session.absolute_url()
+            except ValueError:
+                logger.warn(
+                    "Could not get session URL for session %s and "
+                    "zodb_path %s (session id: %s). There might be a data "
+                    "inconsistency. Not including this session in the "
+                    "notification mailing.",
+                    session.title,
```
ale-rt, line 62: Please do not use `%s`  when substituting what could be a multiline text, use `%r` we were already been asked to do something similar.

ale-rt (changes requested):

src/euphorie/client/notifications/base.py
```diff
@@ -43,12 +47,24 @@ def recipient(self):
...
-            ]
-        )
+
+        session_links = ""
+        for session in self.sessions:
+            session_url = ""
+            try:
+                session_url = session.absolute_url()
+            except ValueError:
+                logger.warn(
```
ale-rt, line 57: Isn't it better to have a:
```suggestion
                logger.exception(
```
How this inconsistency can be fixed?

src/euphorie/client/notifications/base.py
thet: Done. would the title for example be multiline? and how would `__repr__` over `__str__` make a difference?
However, I have changed that.

thet: okay, there's a difference in escaping:

```
>>> "ab %s c" % """okay
... nay"""
'ab okay\nnay c'

>>> "ab %r c" % """okay
... nay"""
"ab 'okay\\nnay' c"
```

src/euphorie/client/notifications/base.py
thet: done.

ale-rt:
I still do get why all of a sudden we have this unconsistency...
Is it a possibility that we have session that do not resolve a proper path?

thet:
@reinhardt I now changed handling of missing titles from:

Screenshot from 2023-12-01 16-06-44

to:

Screenshot from 2023-12-01 16-12-03

PR #680, opened 2023-12-05
Notifications: Don't send mail if recipient has no email address.

thet:
Ref: scrum-1572

### DIFF
--- a/src/euphorie/client/notifications/base.py
+++ b/src/euphorie/client/notifications/base.py
@@ -103,6 +103,15 @@ from OiRA, you can change this [here](${preferences_link})**""",
             )
         )
 
+    def send_email(self):
+        if not self.account.email:
+            logger.warning(
+                "Account %r has no email address. Not sending notification email.",
+                self.account.id,
+            )
+            return
+        super().send_email()
+
 
 class BaseNotification:
     id = None
--- a/src/euphorie/client/notifications/base.py
+++ b/src/euphorie/client/notifications/base.py
@@ -104,7 +104,7 @@ from OiRA, you can change this [here](${preferences_link})**""",
         )
 
     def send_email(self):
-        if not self.account.email:
+        if not self.webhelpers.get_user_email(self.account):
             logger.warning(
                 "Account %r has no email address. Not sending notification email.",
                 self.account.id,


src/euphorie/client/notifications/base.py
```diff
@@ -103,6 +103,15 @@ def index(self):
             )
         )
 
+    def send_email(self):
+        if not self.account.email:
```
reinhardt, line 107: This won't work for DTMB because the email address is not in the account there. On the other hand in plain euphorie this is unnecessary because the user ID is an email address, so every account has an email.

You probably need to use `webhelpers.get_user_email()`.

reinhardt (changes requested):

src/euphorie/client/notifications/base.py
ale-rt: IMO you should not even reach this point.
Results should be filter out earlier.

reinhardt (approved):

src/euphorie/client/notifications/base.py
```diff
@@ -104,7 +104,7 @@ def index(self):
         )
 
     def send_email(self):
-        if not self.account.email:
+        if not self.webhelpers.get_user_email(self.account):
```
ale-rt, line 107: Please do not ignore my comment.
You should not even be here at this point.

src/euphorie/client/notifications/base.py
thet: @reinhardt tnx for the pointer. fixed.

@ale-rt that would involve two things:
a) Implementing this logic and the logging message in all of the different notification classes.
b) or restructuring the code so that the notification base class does that.

I think we can't or shouldn't filter out on the SQL query level - the ws_user_cache reponse JSON‌ field wouldn't allow that easily (not considering Postgres' JSON capabilities, which I‌ don't have experience with) and we would loose the logging feature.

I'm :-1: for a) and :+0: for b) - we're over budget and we would need a full test circle again.

reinhardt: Yes, Alex is going to be furious if we do more than this quick fix.

thet:
Merging this.
@ale-rt we can always optimize later. We might even find the need to rethink the architecture if we find any shortcomings when this feature is used in production.

PR #683, opened 2023-12-21
Use weasyprint for PDF generation

reinhardt:
syslabcom/scrum#732


ale-rt (changes requested):
This is not building for me due to missing pins.
```
Version and requirements information containing pillow:
  [versions] constraint on pillow: 6.2.2
  Requirement of weasyprint: Pillow>=9.1.0
While:
  Installing instance.
Error: The requirement ('Pillow>=9.1.0') is not allowed by your [versions] constraint (6.2.2)
make: *** [Makefile:13: .installed.cfg] Error 1
```

Please before continuing check #603 and rebase on it.

Also take in to account the same notes I did on the other PR (logging, error loading resources and so on).

versions-6.0.cfg
```diff
@@ -40,6 +44,7 @@ path.py = 12.5.0
 perfmetrics = 3.3.0
 pexpect = 4.8.0
 pickleshare = 0.7.5
+Pillow = 9.5.0
```
ale-rt, line 47: This is not needed:
```suggestion
```

Plone 6.0.9 already has the proper version set:
```
ale@avo:~/Code/plone/projects/daimlertruck$ ploneversions 6.0.9|grep Pillow
Pillow = 9.5.0
```

README.rst
```diff
@@ -61,9 +61,8 @@ The configuration is now stored in the registry.
 
 More informations can be found in NuPlone 2 README file.
 
-Euphorie 12 does uses ``wkhtmltopdf`` (and ``xvfb-run`` if available) to render PDF files.
+Euphorie 12 uses ``weasyprint`` to render PDF files.
```
ale-rt, line 64: This sentence should be fixed.
At least update the version number.

reinhardt, line 64:
```suggestion
Euphorie uses ``weasyprint`` to render PDF files.
```

reinhardt:
Is this OK to merge or do you want the log level hack from quaive/ploneintranet#5319 here as well?

ale-rt:
I was not yet able to try this branch,

The goal for me is to check that we have a solution that produces a PDF with a quality comparable with the one wkhtmltopdf was producing and that has no side effects.
I would not like to review this PR again and come back to you asking changes because the PDF differs too much from the previous implementation or because we are flooded by log lines.

So please, check that the PR is in a mergeable state before I invest time in reviewing it.

Thank you.

reinhardt:
> produces a PDF with a quality comparable with the one wkhtmltopdf was producing and that has no side effects

I'm not aware that this is part of a code review. It sounds like you want to do QA already.

ale-rt:
I tested this and the problem of the excessive logging still exists.
In addition, the two produced PDFs look quite a lot different.
image
Left is wkhtmltopdf, right is weasyprint.

I would say it would be ok if they are not identical but IMO the font size is too big for the new report.
This also has the consequence of having the date split on two lines:
image

Can you please recheck?

ale-rt (approved):
I really don't think 8px is the way to go, sorry.
Before we had the possibility to setup options for wkhtmltopdf, and it is a pity we cannot do that anymore.
Anyway, I am merging with my eyes closed.

PR #686, opened 2023-12-22
Add versions.cfg

thet:
When running buildout from another repo which extend the Euphories versions.cfg via raw urls I get this error:
```
daimler@daimler:~/daimler.buildout$ make 
./py3/bin/buildout -Nt 2
While:
  Initializing.
Error: Error downloading extends for URL https://raw.githubusercontent.com/euphorie/Euphorie/main/versions.cfg: HTTP Error 404: Not Found
make: *** [Makefile:11: .installed.cfg] Error 1
```

This PR adds a versions.cfg which extends versions-5.2.cfg to allow buildout run without creating a symlink to a preferred Plone version, so that buildout.cfg can be referenced via GitHub raw URLs.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -1,0 +1,2 @@
+[buildout]
+extends = versions-5.2.cfg


ale-rt (changes requested):
Please fix the parent buildout

thet:
Actually, in the end it wasn't an issue anymore.

In https://github.com/euphorie/euphorie-project which was in use on alita for /home/oira/dtmb there was this change not even necessary.

On the daimler vm 192.168.10.19 on ext0 under /home/daimler/daimler.buildout which is using git@github.com:syslabcom/daimler.buildout.git it was a problem, but that instance is super outdated and I stopped using it.

PR #692, opened 2024-01-26
Sanitize copying a survey to the client.

ale-rt:


mauritsvanrees (approved):
LGTM.  Much clearer and more logical now.
And most importantly: now new tools actually show up on the client side after copying.

PR #693, opened 2024-01-30
Short report: Mark postponed risks

reinhardt:
syslabcom/scrum#1852


src/euphorie/client/docx/compiler.py
```diff
@@ -815,6 +817,9 @@ def set_modules_rows(self, data):
             count = 0
             for risk in risks:
                 answer = risk.get("justifiable", "")
+                postponed = risk.get("postponed", False)
+                if not answer and postponed:
```
ale-rt, line 821: There is no point in getting postponed if answer is falsish, but ok...

src/euphorie/client/docx/compiler.py
```diff
@@ -832,9 +837,12 @@ def set_modules_rows(self, data):
                 count += 1
                 self.set_cell_risk(row_risk.cells[self.risk_description_col], risk)
                 if self.risk_answer_col is not None:
-                    row_risk.cells[self.risk_answer_col].text = (
-                        self.justifiable_map.get(answer) or ""
-                    )
+                    cell = row_risk.cells[self.risk_answer_col]
+                    cell.text = self.justifiable_map.get(answer) or ""
+                    font = self.justifiable_font.get(answer)
```
ale-rt, line 842: I think this increases the complexity of the method, which is already complex. But ok...
Probably adding a separate method, expanded in the `DocxCompilerShort` would have been better.
But I am not able to really judge by just reading the PR.

ale-rt (approved):
I approve, but I suspect some improvements seem desirable.
If you have no time to do that now, please remember that for the future.

reinhardt:
I refactored a little.

Maybe the whole module would benefit from some more refactoring... I'll see if I can sneak something in.

> please remember that for the future

The whole point of code reviews is that a single person cannot think of all possible improvements all the time. Asking me to do this seems both condescending and pointless to me.

ale-rt:
> The whole point of code reviews is that a single person cannot think of all possible improvements all the time. Asking me to do this seems both condescending and pointless to me.

Sorry for that, I thought I was doing good

PR #706, opened 2024-02-26
Refactor getTreeData to use browser view

mauritsvanrees:
Then we can override the browser view for psychosocial sessions.


ale-rt (approved):

ale-rt:
The tests need to be fixed, but I like this philosophy 100%... The more logic we have in views, the easier is to maintain the code

mauritsvanrees:
I have fixed the tests.  I will merge.

PR #707, opened 2024-02-28
Fix locking info in status view

reinhardt:
syslabcom/scrum#2001

### DIFF
--- a/src/euphorie/client/browser/templates/status_info.pt
+++ b/src/euphorie/client/browser/templates/status_info.pt
@@ -90,18 +90,22 @@
           <metal:last_published define-slot="last_published">
             <tal:published condition="webhelpers/use_locking_feature">
               <tal:switch switch="session/review_state">
-                <tal:case case="string:published">
+                <tal:case tal:define="
+                            last_locking_event session/last_locking_event;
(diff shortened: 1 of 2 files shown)


src/euphorie/client/browser/templates/status_info.pt
```diff
@@ -90,18 +90,22 @@
           <metal:last_published define-slot="last_published">
             <tal:published condition="webhelpers/use_locking_feature">
               <tal:switch switch="session/review_state">
-                <tal:case case="string:published">
+                <tal:case tal:define="
+                            last_locking_event session/last_locking_event;
```
ale-rt, line 94: What happens if `last_locking_event` is `None`?

reinhardt: Then `review_state` is `private` and this part is not displayed.

https://github.com/euphorie/Euphorie/blob/scrum-2001-status-locking/src/euphorie/client/model.py#L870

reinhardt: ... but thanks for asking; this is exactly the stuff that often goes untested and causes trouble!

PR #715, opened 2024-03-20
Support scaled answers

mauritsvanrees:
Add support for scaled answers.  Usually: answers on a scale of 1-5.
This includes improvements in the navigation, mostly creating more possibilities for overriding code.
This seems safe enough to be merged and used, also when the specific project does not use the new scaled answer possibilities.
We could also split this up into multiple PRs, and do git rebases, to make this easier to digest.


src/euphorie/content/utils.py
```diff
@@ -426,3 +426,52 @@ def __call__(self):
...
+    result = []
+    if not contents:
+        return result
+    count = 0
+    for answer in contents.splitlines():
+        answer = answer.strip()
+        if not answer:
+            continue
+        count += 1
+        parts = answer.split("|")
```
ale-rt, line 463: I think you want to use `.rpartition` here

mauritsvanrees: Done, looks simpler indeed.

mauritsvanrees:
@ale-rt I made `navigation_tree_legend` a property as you requested.  I have pushed the same change to the `daimler.oira` branch.

ale-rt:
> @ale-rt I made `navigation_tree_legend` a property as you requested. I have pushed the same change to the `daimler.oira` branch.

I meant something like this: https://github.com/euphorie/Euphorie/pull/716
I moved that property to an attribute.
It can still be customized with a property.

mauritsvanrees:
Ah, sorry, now I get it.  Makes sense.  I have merged your PR.

src/euphorie/content/utils.py
ale-rt: Please look at this https://github.com/euphorie/Euphorie/pull/717, it also cover another edge case

mauritsvanrees:
@ale-rt and me are splitting this up into multiple PRs so it is more understandable.

ale-rt:
Almost, I merged main in the multiple answer branch and now I have only this thing left:
```
--- a/src/euphorie/client/browser/risk.py
+++ b/src/euphorie/client/browser/risk.py
@@ -1245,7 +1245,9 @@ class ActionPlanView(RiskBase):
 
     @property
     def risk_postponed(self):
-        return self.context.identification is None
+        return (
+            self.context.identification is None and self.context.scaled_answer is None
+        )
 
     @property
     def use_problem_description(self):
```

ale-rt:
This was removed on purpose :)

PR #719, opened 2024-04-16
Delete guest account after sessions have been transferred.

reinhardt:
Touch the sessions before that to remove the reference to the guest account in `last_modifier_id` column.

Fixes syslabcom/scrum#2155

### DIFF
--- a/src/euphorie/client/browser/login.py
+++ b/src/euphorie/client/browser/login.py
@@ -105,15 +105,16 @@ class Login(BrowserView):
             return
 
         old_authenticated_account_id = old_credentials[0]
-
-        if (
+        old_account = (
             Session.query(model.Account)
             .filter(
                 model.Account.account_type == config.GUEST_ACCOUNT,
                 model.Account.id == old_authenticated_account_id,
             )
             .first()
-        ) is None:
+        )
+
+        if old_account is None:
             # We check that the previously authenticated user was actually
             # a guest account that has been converted.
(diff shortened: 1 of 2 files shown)


src/euphorie/client/browser/login.py
```diff
@@ -126,6 +127,8 @@ def transferGuestSession(self):
         )
         for session in sessions:
             session.account_id = new_account.id
+            session.touch()
```
ale-rt, line 130: Calling `.touch` does two things:

1. sets the last modifier
2. updates the section modified time

See https://github.com/euphorie/Euphorie/blob/af3e98fdae4b199cb6962e8e2c75d0dff6e14aa4/src/euphorie/client/model.py#L881-L883

While you want the first, I am not really sure the second is needed.

I wonder if we just can do:
```
session.last_modifier_id = new_account.id
```
without  changing the modification date of the session.

ale-rt (approved):
Fine to merge, but please see my comment

src/euphorie/client/browser/login.py
reinhardt: None of this is “needed”. That's missing the point. I know there are at least four different ways of handling this. It's hard to decide if you don't tell me any reason to do it differently.

ale-rt: Discussed on a call, we were not able to agree.
@pilz Please jump in the discussion

PR #722, opened 2024-04-19
client navigation: refactor getTreeData function to call a browser vi…

mauritsvanrees:
…ew 'oira_navigation_tree'.

Then we can subclass this browser view elsewhere to change the tree in one spot, instead of needing to override the tree in multiple views. Taken from https://github.com/euphorie/Euphorie/pull/715


ale-rt (approved):

mauritsvanrees:
First we need to approve and merge PR #724 with the scaled answers.

ale-rt (approved):

PR #724, opened 2024-04-19
Support scaled answers.

mauritsvanrees:
These are answers on a scale from usually 1-5, instead of only yes/no. This is the main part from https://github.com/euphorie/Euphorie/pull/715. We are pulling that apart into some more digestible chunks, instead of 50 commits, including merges of main.

I wonder if this needs an upgrade step, because in `model.py` we add a `scaled_answer` column.  We have code for that in a customer-specific project.


ale-rt:
Yes this needs an upgrade step

PR #725, opened 2024-04-23
Clean up references to guest accounts in last_modifier_id

reinhardt:
In acc2353126 (#719) we changed `transferGuestSession` to update `last_modifier_id` and delete the guest account after all sessions have been transferred. This upgrade step does a similar thing with existing sessions. We don't update the modification date because we don't know when the session was transferred. We also double check that the guest account doesn't have any sessions associated with it so that we can safely delete it.

Fixes syslabcom/scrum#2155


thet (approved):
LGTM

PR #726, opened 2024-04-26
Scrum 2193  notifications allow user settings

thet:
Ref: scrum-2193

Define a registry record to dissallow users to set their notification settings in the user profile, while still showing what is configured. Also show a info explaining why this cannot be changed.

image

For the speficic site, this setting can simple be set via the registry.


reinhardt (commented):
As stated in the ticket I disagree with the premise, but in terms of code quality it's just fine.

PR #731, opened 2024-05-08
Show training certificates from other users in my organisation

reinhardt:
The training certificate view was previously meant to only show the current user's certificates. Now it can take a parameter `training_id` that makes it display the training with the given id instead. The default is still to show the current user's certificate connected to the respective session (there can be only one such certificate per session). The parameter can also be passed to the `index()` method so that we don't have to fiddle with the request when displaying multiple certificates like in the overview page.

Ref syslabcom/scrum#2142


reinhardt:
Note to self: Process `training_id` in print view etc.

cillianderoiste (approved):

PR #741, opened 2024-05-20
The tests folder must be a module for the tests to be discovered

ale-rt:


mauritsvanrees (approved):
Ah right, this was missing, thanks.

PR #747, opened 2024-05-27
Improve the widget for the sector logo

ale-rt:
Allows to remove a customized template that rewrites the whole form in other packages, see for example:

- https://github.com/euphorie/osha.oira/blob/31dec4fd41d554fcce429a4254a1935aa7493bad/src/osha/oira/content/browser/templates/sector_edit.pt#L32-L63

But we have something similar for other projects.

### DIFF
--- a/src/euphorie/content/widgets/logo.py
+++ b/src/euphorie/content/widgets/logo.py
@@ -1,0 +1,10 @@
+from plone.formwidget.namedfile.widget import NamedImageWidget
+from z3c.form.widget import FieldWidget
+
+
+class LogoWidget(NamedImageWidget):
+    pass
+
+
+def LogoFieldWidget(field, request):
+    return FieldWidget(field, LogoWidget(request))
--- a/src/euphorie/content/widgets/templates/logo.pt
+++ b/src/euphorie/content/widgets/templates/logo.pt
@@ -1,0 +1,63 @@
+<fieldset class="concise">
+  <p class="legend"
+     i18n:translate="label_logo"
+  >Logo</p>
+  <dfn class="infoPanel"
+       title="Information"
+       i18n:attributes="title"
+       i18n:translate="help_sector_logo"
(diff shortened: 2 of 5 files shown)


src/euphorie/content/widgets/logo.py
```diff
@@ -0,0 +1,10 @@
+from plone.formwidget.namedfile.widget import NamedImageWidget
+from z3c.form.widget import FieldWidget
+
+
+class LogoWidget(NamedImageWidget):
+    pass
```
thet, line 6: You could limit the accepted types via the `accept` attribute, if you want.
e.g.:
```
class LogoWidget(NamedImageWidget):
    accept = ("*.png", "*.jpg", "*.gif",  "*.webp")
```
... or maybe add the accept parameter in the directives statement above.

Note: out of my head, untested....

See: https://github.com/plone/plone.namedfile/pull/158

src/euphorie/content/widgets/templates/logo.pt
```diff
@@ -0,0 +1,63 @@
+<fieldset class="concise">
+  <p class="legend"
+     i18n:translate="label_logo"
+  >Logo</p>
+  <dfn class="infoPanel"
+       title="Information"
+       i18n:attributes="title"
+       i18n:translate="help_sector_logo"
+  >The logo will appear on the client side app that your user group will see. Make sure your image is of format png, jpg or gif and does not contain any special characters. The new logo will only become visible after you've saved your changes and published the OiRA tool.</dfn>
```
thet, line 9: I'd also recommend to use `webp` as it produces very good results with a lower filesize and also offers lossless compression.

thet (approved):
LGTM, but two remarks I'd look into.

ale-rt:
What you propose are different tasks that should be done in other PR.
This PR has the only one and clear focus of getting rid of the custom template in osha.oira.

PR #758, opened 2024-08-02
Fix Dutch translations around publishing.

mauritsvanrees:
* 'publishing' and 're-publishing' were switched around.  I noticed this because I was working on a fix in this area, and the text was not logical.
* A few translations used the wrong abbreviation of 'RI&E' (Risico Inventarisatie & Evaluatie).
* One translation was missing.
* Plus two typos fixed.

Take a survey that was published already and has some structural changes.
Screen shot before:

Screenshot 2024-08-02 at 14 45 04

Screen shot with this PR:

Screenshot 2024-08-02 at 14 45 55

@pilz Does this need checking by an actual Dutch client?
For me it is fine to merge.  But no hurry.


ale-rt:
This might be related to the fact that the Nederlands have a custom package (https://github.com/euphorie/tno.euphorie/)

PR #763, opened 2024-09-25
Similar titles: Store HTML as annotation until new search is triggered

reinhardt:
Note that the minimum and maximum similarity values are not stored. When you call the view the default values are displayed (0.5 and 1.0, resp.) but they may not match the results that are shown. I would say that's OK for now, but could be a possible future improvement.

Ref syslabcom/scrum#2517

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -18,6 +18,10 @@ Changelog
   Ref: scrum-2669
   [reinhardt]
 
+- Similar titles: Store results until explicit refresh
+  Ref: scrum-2517
+  [reinhardt]
+
 
 16.2.1 (2024-07-31)
 -------------------
--- a/src/euphorie/content/browser/configure.zcml
+++ b/src/euphorie/content/browser/configure.zcml
@@ -470,6 +470,15 @@
       />
 
   <browser:page
+      name="similar-titles-stored"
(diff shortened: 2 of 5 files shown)


ale-rt (changes requested):
Nice job, but this will overwrite the results for every search, which is not ideal.

Please add a third button to display the results set in the annotations (if present).
The file has to be set by a cron job nightly.

If it is unclear, let's have a call.

reinhardt:
Is this any closer to what you have in mind? https://github.com/euphorie/Euphorie/pull/763/commits/441628f3a1f3bdcecd0aa09f5008465614a4e560

ale-rt:
Yes, closer. 

This still overwrites the results for every search that a user makes.
Do not do that, please.

Also, I would like to keep the behavior we had before when clicking the search button: the results should be displayed below the form.

Thanks a lot

reinhardt:
Another try: https://github.com/euphorie/Euphorie/pull/763/commits/6823104ef88523b68971eb4f2313658e34e78a3b

The cron job will need to make a post request, or we can make a shortcut view or something that calls `similar-titles-stored` internally.

ale-rt (approved):

PR #766, opened 2024-10-28
Fix nl-BE translation

reinhardt:
Ref syslabcom/scrum#2769


ale-rt (approved):
I approve... but the translation consistency looks improvable

PR #768, opened 2024-11-06
ActionPlanTimeline: Strip whitespace when converting to plain text

reinhardt:
Follow-up for #767 

Ref syslabcom/scrum#2763


ale-rt (changes requested):
See https://github.com/euphorie/osha.oira/pull/289#discussion_r1830976301

ale-rt (approved):

PR #770, opened 2024-11-20
Fix query for unanswered risks

reinhardt:
Ref syslabcom/scrum#2800

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -10,6 +10,9 @@ Changelog
 - Action Plan: Strip HTML from comments
   (`#2763 <https://github.com/syslabcom/scrum/issues/2763>`_)
   [reinhardt]
+- Report: Fix “parked” risks
+  (`#2800 <https://github.com/syslabcom/scrum/issues/2800>`_)
+  [reinhardt]
 
 
 16.2.3 (2024-10-25)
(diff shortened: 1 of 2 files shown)


ale-rt (approved):
I have no idea what are the implication of this change for other projects that use Euphorie.

Anyway, I see that we have more or less the same statement here:
https://github.com/euphorie/osha.oira/blob/4d34265ee9383964ff3c095e92ce48cb4a3cf539/src/osha/oira/client/model.py#L166-L178

Is that just copy-pasted/dead code?

I approve based on trust, but I admit this change would take a lot more time to be reviewed, especially because of the implications on other projects.

The lack of tests for sure does not help.

reinhardt:
> I have no idea what are the implication of this change for other projects that use Euphorie.

I'm reasonably sure that this is a sensible fix for eveyone. The query is supposed to return unanswered risks, as the name suggests, but it was returning many answered (identified) risks as well, which were then duplicated in the report.

> Anyway, I see that we have more or less the same statement here: https://github.com/euphorie/osha.oira/blob/4d34265ee9383964ff3c095e92ce48cb4a3cf539/src/osha/oira/client/model.py#L166-L178
> 
> Is that just copy-pasted/dead code?

Thanks for pointing that out, I didn't know that. I'll look into it!

PR #773, opened 2024-11-22
Fix query for risks that are not present

reinhardt:
Analogous to #770 

Ref syslabcom/scrum#2800


ale-rt (approved):
LGTM but pre-commit complains about an import that is not needed anymore

PR #779, opened 2025-01-07
Support openpyxl 3.x

reinhardt:


src/euphorie/client/browser/report.py
```diff
@@ -252,4 +249,9 @@ def __call__(self):
             "Content-Type",
             "application/vnd.openxmlformats-" "officedocument.spreadsheetml.sheet",
         )
-        return save_virtual_workbook(book)
+
+        with NamedTemporaryFile(delete=False) as tmp:
```
ale-rt, line 253: It seems strange to me that we have delete False.

ale-rt (commented):
Is this to support Python 3.11?
In this case please also add the proper GHA tests

src/euphorie/client/browser/report.py
reinhardt: It's modelled after this example: https://docs.python.org/3/library/tempfile.html#examples

Note that on python 3.13 the parameter is `delete_on_close` which is a more accurate name.

src/euphorie/client/browser/report.py
```diff
@@ -252,4 +249,9 @@ def __call__(self):
...
+            save_workbook(book, tmp.name)
+            tmp.close()
+            with open(tmp.name, mode="rb") as f:
+                return f.read()
```
ale-rt, line 257: tmp.seek(0)
```suggestion
            tmp.seek(0)
            return f.read()
```

src/euphorie/client/browser/report.py
reinhardt: The context manager is doing that. The parameter only prevents deletion when `close()` is called. When the context is exited the file is deleted anyway.

The documentation is a bit ambiguous, but that's how I understand it.

> If _delete_ is true (the default), the file is deleted as soon as it is closed.

https://docs.python.org/3.11/library/tempfile.html#tempfile.NamedTemporaryFile

In 3.13 there is an improvement:

> If _delete_ is true and _delete_on_close_ is false, the file is deleted on context manager exit only

https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

reinhardt: Meh... I tested it - the file is not cleaned up by the context manager. I'll change the code.

src/euphorie/client/browser/report.py
reinhardt: It feels right to me, because we're first writing and then reading. That's different modes.

ale-rt:
@reinhardt why did you merge this, it was not approved and I asked for changes...

src/euphorie/client/browser/report.py
```diff
@@ -218,11 +219,7 @@ def create_workbook(self):
                             value = module.title
                 if value is not None:
                     cell = sheet.cell(row=row, column=column)
-                    if key == "number":
-                        # force sting
-                        cell.set_explicit_value(value)
```
ale-rt, line 223: I think this was needed.

reinhardt:
GHA tests for python 3.11 are here: https://github.com/euphorie/Euphorie/pull/780

> https://github.com/euphorie/Euphorie/pull/779#discussion_r1905455627

This was added after the PR was merged. That said, I can look into it again, but I doubt that it's needed. The method `set_explicit_value` doesn't even exist in newer openpyxl versions. It used to allow a `data_type` parameter, but we're not even passing that.

See https://foss.heptapod.net/openpyxl/openpyxl/-/blob/1a12044865c82190bdfb39b1c02b7411b3f52d00/openpyxl/cell/cell.py#L254

ale-rt:
> GHA tests for python 3.11 are here: #780
> 

They are failing.

> > [#779 (comment)](https://github.com/euphorie/Euphorie/pull/779#discussion_r1905455627)
> 
> This was added after the PR was merged. That said, I can look into it again, but I doubt that it's needed. The method `set_explicit_value` doesn't even exist in newer openpyxl versions. It used to allow a `data_type` parameter, but we're not even passing that.
> 

That means we use the default.

> See https://foss.heptapod.net/openpyxl/openpyxl/-/blob/1a12044865c82190bdfb39b1c02b7411b3f52d00/openpyxl/cell/cell.py#L254

I already checked that yesterday.
What is your point?

reinhardt:
We don't need the case distiction.

PR #781, opened 2025-01-08
Add view that lists all links in a tool

reinhardt:
Ref https://github.com/syslabcom/scrum/issues/2757

### DIFF
--- a/src/euphorie/content/browser/survey.py
+++ b/src/euphorie/content/browser/survey.py
@@ -42,6 +42,8 @@ from zope.container.interfaces import INameChooser
 from zope.event import notify
 from zope.i18n import translate
 
+import re
+
 
 class SurveyBase(BrowserView):
     def _morph(self, child):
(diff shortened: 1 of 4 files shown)


src/euphorie/content/browser/survey.py
```diff
@@ -520,3 +522,51 @@ def tools(self):
...
+    attributes_checked = [
+        "description",
+        "introduction",
+        "solution_direction",
+        "legal_reference",
+        "action",
+        "action_plan",
+        "prevention_plan",
+        "requirements",
+    ]
```
mauritsvanrees, line 537: Are you sure these are all the fields in all relevant content types?
You could dynamically get all text fields and iterate over them.  See how [collective.searchandreplace does it](https://github.com/collective/collective.searchandreplace/blob/master/collective/searchandreplace/searchreplaceutility.py#L247-L249).  Ignore the Archetypes-related code. ;-)

src/euphorie/content/browser/survey.py
```diff
@@ -520,3 +522,51 @@ def tools(self):
...
+    url_regex = re.compile(
+        r"https?://[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b"
+        r"(?:[-a-zA-Z0-9()@:%_+.~#?&/=]*)"
+    )
```
mauritsvanrees, line 541: This is probably fine.  Alternative idea: look for contents of href/src, so `="http..."`, with at the dots anything except a double quote or a space.  Something like this:

```
>>> re.compile(r'="(https?://[^ "]*)"').match('="https://some.url"').groups()
('https://some.url',)
```

src/euphorie/content/browser/survey.py
```diff
@@ -520,3 +522,51 @@ def tools(self):
...
+            value = getattr(aq_base(obj), attrib, "")
+            if value:
+                links.extend(self.url_regex.findall(value))
+        if links:
+            yield {
+                "object": obj,
+                "links": [{"url": link, "text": link} for link in links],
+            }
+        if hasattr(obj, "objectValues"):
+            for child in obj.objectValues():
```
mauritsvanrees, line 555: Theoretically `objectValues` can contain non-contentish items, for example an object for local workflow policies.  If I want only content I prefer `contentValues`:

```suggestion
        if hasattr(obj, "contentValues"):
            for child in obj.contentValues():
```

src/euphorie/content/browser/templates/list-links.pt
```diff
@@ -0,0 +1,35 @@
...
+
+  <metal:title fill-slot="title">List of Links in ${context/title}</metal:title>
+  <metal:content fill-slot="content">
+    <article>
+      <dl>
+        <tal:item tal:repeat="item view/items">
+          <dt tal:define="
+                obj python:item['object'];
+              ">
+            <a href="${obj/absolute_url}">${obj/title}</a>
```
mauritsvanrees, line 21: As you suggest in the issue, I would say it is better to link to the edit form directly:

```suggestion
            <a href="${obj/absolute_url}/edit">${obj/title}</a>
```

mauritsvanrees (commented):
I made some suggestions.  But it seems to work when I try it locally.

pilz:
@reinhardt thank you for the work, works fine!
@gyst is working on the link checker (#846) which bases on this PR, so this will be included when we merge his.
@mauritsvanrees thank you for the suggestions. The contentViews and the /edit I took over into Guidos PR, the general approach from c.searchandreplace and the change in the regex I shyed away due to time constraints, but we might revisit.

pilz:
merged as part of the link checker effort.

PR #783, opened 2025-01-14
Don't update titles of session-specific instances of profilequestions

reinhardt:
...from ZODB

Ref https://github.com/syslabcom/scrum/issues/3038

### DIFF
--- a/src/euphorie/client/model.py
+++ b/src/euphorie/client/model.py
@@ -902,6 +902,12 @@ class SurveySession(BaseObject):
                 if item.zodb_path.find("custom-risks") >= 0:
                     continue
                 zodb_item = survey.restrictedTraverse(item.zodb_path.split("/"), None)
+                # Don't update session-specific instances of profilequestions
+                if (
+                    zodb_item.portal_type == "euphorie.profilequestion"
+                    and item.profile_index > -1
+                ):
+                    continue
                 if zodb_item and zodb_item.title != item.title:
                     item.title = zodb_item.title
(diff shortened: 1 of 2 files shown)


src/euphorie/client/model.py
```diff
@@ -902,6 +902,12 @@ def refresh_survey(self, survey=None):
                 if item.zodb_path.find("custom-risks") >= 0:
                     continue
                 zodb_item = survey.restrictedTraverse(item.zodb_path.split("/"), None)
+                # Don't update session-specific instances of profilequestions
+                if (
+                    zodb_item.portal_type == "euphorie.profilequestion"
+                    and item.profile_index > -1
```
thet, line 908: I have no idea what this profile_index is and it seems not be explained anywhere in model.py and why having it bigger than "-1"‌ indicates that it is session-specific.
A little comment would help :)

thet (approved):
LGTM, except that I don't know about the business logic. See comment.

Feel free to merge. If you add a comment what this profile_index > -1 is about, I'd be happy.

src/euphorie/client/model.py
reinhardt: Yeah, we'd probably have to fly in Wolfgang for that. If you grep for `profile_index` you find some examples - that was all I had to go by. But I suspect that this feature is not currently used in practice anyway. This code was last updated five years ago and nobody noticed the problem until Angel happened to check a test tool that has profile questions activated.

PR #791, opened 2025-01-24
Custom labels in report_company for France

reinhardt:
We're not going through the translation mechanism here because other French speaking countries (Belgium) should not be affected.

Ref https://github.com/syslabcom/scrum/issues/2375


thet (approved):
LGTM

PR #792, opened 2025-01-28
Update prototype from commit da8a85739b14e3d8791cecc9e89979f5f643def7

pilz:
Resource update from prototype to prepare for usability testing.


ale-rt:
I think this should be considered obsolete.
The package to update is now https://github.com/euphorie/plonestatic.euphorie

PR #793, opened 2025-01-28
Markup alignment for usability testing

pilz:
I did a session with Daniel today and aligned markup with proto so that we are better prepared for the upcoming usability testing. 
These changes should be merged together with the PR to update the resources to the latest version - see https://github.com/euphorie/Euphorie/pull/792 (but is not directly coupled). The fixes are safe, and also work with an older version. Also, the proto update is safe, we clicked through, there was no visible breakage.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -19,6 +19,9 @@ Changelog
   (`#2375 <https://github.com/syslabcom/scrum/issues/2375>`_)
   [reinhardt]
 
+- Updated resources from proto and applied a number of markup fixes in a review session with Daniel
+  [pilz]
+
 
 16.2.7 (2025-01-15)
 -------------------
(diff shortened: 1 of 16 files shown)


pilz:
@ale-rt I think I know exactly what the problem is but I fail to understand how I can run single tests locally. Can you tell me the command? (bin/test -t /-m etc failed with 0 tests ran)

ale-rt:
> @ale-rt I think I know exactly what the problem is but I fail to understand how I can run single tests locally. Can you tell me the command? (bin/test -t /-m etc failed with 0 tests ran)

If you use the buildout included in this package, according to:

https://github.com/euphorie/Euphorie/blob/9ab904821d2a42a85099fc68c88bbd5f89b523ee/.github/workflows/main.yml#L50-L53

you should be able to run the tests with `./bin/test -s euphorie`.

pilz:
Yes, but this always runs all tests. Can't I run just a subset to save time? @ale-rt

ale-rt:
> Yes, but this always runs all tests. Can't I run just a subset to save time? @ale-rt

Of course, passing `-t test_name`.
Ping me in chat because it is not easy to guide you from here.

ale-rt (approved):

PR #797, opened 2025-02-03
The bundle URL changed for whatever reason

ale-rt:


thet (approved):
:+1: 

I removed the "/js" - the reason was that I don't see something else than the Patternslib JavaScript being provided by plone.patternslib.

PR #799, opened 2025-02-04
Short report: Only show Measures in Place heading if tool supports it

reinhardt:
Ref https://github.com/syslabcom/scrum/issues/3090


ale-rt (approved):
LGTM

PR #801, opened 2025-02-14
Identification: Split up call method for easier overrides

reinhardt:
The diff is a bit hard to read, but this only moves most of the code in `__call__()` out to `post()`, and then in turn a big part of that to `save_data()`.

The idea is that we can create a subclass that has a custom `post()` method which does not call `save_data()` but retains the rest of the functionality.

Ref syslabcom/scrum#2867

### DIFF
--- a/src/euphorie/client/browser/risk.py
+++ b/src/euphorie/client/browser/risk.py
@@ -509,6 +509,61 @@ class IdentificationView(RiskBase):
             return None
         return condition
 
+    def save_data(self, reply):
+        old_values = {}
+        for prop, default in self.monitored_properties.items():
+            val = getattr(self.context, prop, default)
+            old_values[prop] = val
+
+        self.set_answer_data(reply)
+
+        session = Session()
+        changed = self.set_measure_data(reply, session)
+
+        if reply.get("answer", None):
+            # If answer is not present in the request, do not attempt to set
+            # any action-related data, since the request might have come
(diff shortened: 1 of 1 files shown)


src/euphorie/client/browser/risk.py
```diff
@@ -509,6 +509,61 @@ def action_plan_condition(self):
...
+        if not self.webhelpers.can_edit_session:
+            return self.proceed_to_next(reply)
+        _next = self._get_next(reply)
+        # Don't persist anything if the user skipped the question
+        if _next == "skip":
+            return self.proceed_to_next(reply)
+
+        self.save_data(reply)
+
+        return self.proceed_to_next(reply)
```
thet, line 565: This would untangle the code a bit for the cost of not being a simple restructuring anymore.
Just ignore this.

```suggestion
        if self.webhelpers.can_edit_session:
            _next = self._get_next(reply)
            # Only persist, if the user hasn't skipped the question
            if _next != "skip":
                self.save_data(reply)

        return self.proceed_to_next(reply)
```

thet (approved):
LGTM

src/euphorie/client/browser/risk.py
reinhardt: I did consider this, but went for only splitting up the methods for now. Maybe I'll get around to it while working on the DS etool.

PR #804, opened 2025-02-17
Profilequestion: Change question field to rich text

reinhardt:
Ref syslabcom/scrum#2859

### DIFF
--- a/src/euphorie/content/browser/templates/profilequestion_view.pt
+++ b/src/euphorie/content/browser/templates/profilequestion_view.pt
@@ -24,10 +24,8 @@
 
       <h2 i18n:translate="header_information">Information</h2>
 
+      <p tal:replace="structure python:view.get_safe_html(context.question)"></p>
       <dl class="grid span-13">
-        <dt i18n:translate="label_profilequestion_question">Question</dt>
-        <dd>${context/question}</dd>
-
         <tal:use_location_question condition="context/use_location_question">
           <dt i18n:translate="">Multiple Item Question</dt>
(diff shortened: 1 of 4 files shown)


src/euphorie/content/profilequestion.py
```diff
@@ -38,14 +39,14 @@ class IProfileQuestion(model.Schema, IRichDescription, IBasic):
     be skipped, or repeated multiple times.
     """
 
-    question = schema.TextLine(
+    question = HtmlText(
```
thet, line 42: :eyes: I'd have used `plone.app.textfield.RichText` instead and use Plone's portal_transform machinery with maybe a custom transform and also benefit from transform caching, etc.
But this is the way we do it in euphorie and we don't change that for this PR, alright.

src/euphorie/content/browser/templates/profilequestion_view.pt
```diff
@@ -24,10 +24,8 @@
 
       <h2 i18n:translate="header_information">Information</h2>
 
+      <p tal:replace="structure python:view.get_safe_html(context.question)"></p>
       <dl class="grid span-13">
-        <dt i18n:translate="label_profilequestion_question">Question</dt>
-        <dd>${context/question}</dd>
```
thet, line 29: We could also do something like:

```
        <dd tal:replace="structure python:view.get_safe_html(context.question)"></dd>
```

thet (approved):
Suggestions, but approved.
Just ignore the suggestions, as they are probably not applicable.

src/euphorie/content/profilequestion.py
reinhardt: Yeah, I just want to be consistent for the moment. But maybe we can talk about switching if you see benefits there...

PR #808, opened 2025-02-19
3181 markup fixes

pilz:
Refs https://github.com/syslabcom/scrum/issues/3181 and fixes Angels blockers for a release.

### DIFF
--- a/src/euphorie/client/browser/templates/webhelpers.pt
+++ b/src/euphorie/client/browser/templates/webhelpers.pt
@@ -599,7 +599,7 @@
                       value context/priority;
                     "
           >
-            <legend><tal:span i18n:translate="header_risk_priority">Select the priority of this risk</tal:span></legend>
+            <legend i18n:translate="header_risk_priority">Select the priority of this risk</legend>
             <label><input checked="${python:'checked' if value=='low' else None}"
                      name="priority"
                      type="radio"
--- a/src/euphorie/client/browser/templates/webhelpers.pt
+++ b/src/euphorie/client/browser/templates/webhelpers.pt
@@ -356,8 +356,9 @@
          data-pat-depends="${view/evaluation_condition}; transition: slide"
          tal:define="
            evaluation_method python:risk.type in ('policy', 'top5') and risk.type or risk.evaluation_method;
+           skip_evaluation view/skip_evaluation;
          "
-         tal:condition="not:view/skip_evaluation"
+         tal:condition="python: not skip_evaluation and evaluation_method != 'policy'"
     >
       <p class="problem-description"><strong>${view/risk/problem_description}</strong></p>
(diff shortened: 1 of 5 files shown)


src/euphorie/client/browser/templates/webhelpers.pt
```diff
@@ -351,13 +351,14 @@
...
-    <div class="risk-module pat-well notice pat-depends vertical evaluation"
+    <div class="risk-module pat-well notice pat-depends evaluation"
          id="evaluation"
          data-pat-depends="${view/evaluation_condition}; transition: slide"
          tal:define="
            evaluation_method python:risk.type in ('policy', 'top5') and risk.type or risk.evaluation_method;
+           skip_evaluation view/skip_evaluation;
          "
-         tal:condition="not:view/skip_evaluation"
+         tal:condition="python: not skip_evaluation and evaluation_method != 'policy'"
```
reinhardt, line 361: This seems to be more than a markup change. Why is this necessary?

src/euphorie/client/browser/templates/webhelpers.pt
```diff
@@ -368,250 +369,256 @@
...
-            />
-            <tal:span i18n:translate="priority_high">High</tal:span></label>
+        <fieldset class="vertical">
+          <fieldset class="pat-checklist"
+                    disabled="${python:'disabled' if not webhelpers.can_edit_session else None}"
+                    tal:define="
+                      value context/priority;
+                    "
+          >
+            <legend><tal:span i18n:translate="header_risk_priority">Select the priority of this risk</tal:span></legend>
```
reinhardt, line 602:
```suggestion
            <legend i18n:translate="header_risk_priority">Select the priority of this risk</legend>
```

reinhardt: Why the extra `<tal:span>`?

src/euphorie/client/browser/templates/webhelpers.pt
pilz: Because for policy risks there is no evaluation and then we have an empty box with only a title that should not be there. Pointed out by Daniel as a "Schmutzeffekt", but actually needs a code change.

src/euphorie/client/browser/templates/webhelpers.pt
pilz: Yes, I just replicated the markup from the sections above where there was more in the legend than just text and therefore a tal:span was used. I am fine if you want to change that.

src/euphorie/client/browser/templates/webhelpers.pt
reinhardt: Ah OK. I was afraid we might hide some logic change in a markup PR, but if it's about an empty box it should be fine.

PR #811, opened 2025-02-24
report_company: remove span tag

reinhardt:
Fixes erroneous line breaks

Ref syslabcom/scrum#3181


ale-rt (approved):
I did not check but it LGTM

PR #815, opened 2025-03-10
Add a tool_type_data property in the risk view class

ale-rt:
While this is not directly used in Euphorie code, it is convenient for customers code.


ale-rt:
This will be used in OiRA creator, refs. https://github.com/syslabcom/scrum/issues/3059

reinhardt (approved):

PR #821, opened 2025-03-25
Clean up set_parameter_values

ale-rt:
The parameters that are set are now lazily evaluated when needed.

This removes some duplicated code.

This will be used in OiRA creator, refs. https://github.com/syslabcom/scrum/issues/3059


pilz (approved):
Deprecation and memoize, no complicated things. I don't see any objections :)

PR #818, opened 2025-03-20
Try Plone 6.1

ale-rt:

PR #822, opened 2025-03-25
Use the plonetheme.nuplone.z3cform.widget.WysiwygWidget if available

ale-rt:


ale-rt (commented):
I am merging this because after #818 this is just a changelog

PR #827, opened 2025-03-30
#3385 Create organisation by default

pilz:
This has been urgently requested by Frank and Lia. 
We tested in the usability test.
Having it reduces a lot of distractions for the users.

### DIFF
--- a/src/euphorie/client/browser/organisation.py
+++ b/src/euphorie/client/browser/organisation.py
@@ -133,6 +133,7 @@ class OrganisationBaseView(BaseView):
             organisation = Organisation(
                 owner_id=account.id, title=self.default_organisation_title
             )
+            alsoProvides(self.request, IDisableCSRFProtection)
             self.sqlsession.add(organisation)
             account.organisation = organisation
 
--- a/src/euphorie/client/browser/organisation.py
+++ b/src/euphorie/client/browser/organisation.py
@@ -129,9 +129,14 @@ class OrganisationBaseView(BaseView):
             return []
 
         organisations = []
-        if account.organisation:
-            # Add the user own organisation (if present)
-            organisations.append(account.organisation)
+        if not account.organisation:
+            organisation = Organisation(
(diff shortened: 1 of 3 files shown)


src/euphorie/client/browser/organisation.py
```diff
@@ -129,9 +129,14 @@ def organisations(self):
             return []
 
         organisations = []
-        if account.organisation:
-            # Add the user own organisation (if present)
-            organisations.append(account.organisation)
+        if not account.organisation:
```
ale-rt, line 132: Is this a write on read? Doesn't plone.protect prevent yopu to do this?

ale-rt (commented):
I do not mind if you display a page that acts like the organisation is already present, but you should better delay the creation when you will are adding a user or will edit the organisation.
I verified that plone.protect does not kick in even if you have a write on read, which is strange.
If a write on read is kept, we should have an `alsoProvides(self.request, IDisableCSRFProtection)`

src/euphorie/client/browser/organisation.py
ale-rt: It should bark anyway, see https://github.com/EU-OSHA/oira.application.buildout/pull/8

pilz:
The error is:
Error: getCacheEntry failed: This legacy service is shutting down, effective April 15, 2025. Migrate to the new service ASAP. For more information: https://gh.io/gha-cache-sunset

@ale do you know what needs to be done here?

pilz:
It was the pre-commit that needs to go from 2.0.0 to 3.0.1 to use the new cache api.

ale-rt:
> Should we better change the default view to mimic an organisation and then create it only when a user edits it or tries to add a user?

I think it would be better, but I am fine also to keep it like it is.

ale-rt (approved):

angeldasangel:
Clarified with Ale and informed Lia.

PR #828, opened 2025-03-30
#3384 shortcut create

pilz:
When on the tools overview, there is no need to have the selection modal, as users click the button from the tool tile. Therefore create immediately.


ale-rt (approved):
Approved, note that the pointer does not change when you hover the button, which IMO is a bug as well.

reinhardt (approved):

pilz:
Agreed.

PR #829, opened 2025-03-30
#3384 hide exit on dashboard

pilz:
When the menu is opened from the dashboard, the exit action makes no sense (we only exit sessions). Added a check so that we can hide it.

### DIFF
--- a/src/euphorie/client/browser/templates/more_menu.pt
+++ b/src/euphorie/client/browser/templates/more_menu.pt
@@ -78,6 +78,7 @@
           <a class="pat-inject close-panel icon-cancel"
              href="${webhelpers/country_or_client_url}#content"
              data-pat-inject="history: record"
+             tal:condition="python:'from_dashboard' not in context.REQUEST.keys()"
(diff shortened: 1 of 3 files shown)


src/euphorie/client/browser/templates/more_menu.pt
```diff
@@ -78,6 +78,7 @@
           <a class="pat-inject close-panel icon-cancel"
              href="${webhelpers/country_or_client_url}#content"
              data-pat-inject="history: record"
+             tal:condition="python:'from_dashboard' not in context.REQUEST.keys()"
```
ale-rt, line 81:
```suggestion
             tal:condition="python:'from_dashboard' not in request"
```
Maybe it would be better to check the referrer.

pilz:
I did it using the referer, made a new PR to keep the commits clean. See https://github.com/euphorie/Euphorie/pull/830.
Closing this one.

reinhardt:
FWIW you can force push to a PR branch. Github will detect it and offer to show a diff. @pilz

PR #831, opened 2025-04-01
Old cache api was sundowned today by github

pilz:
see https://gh.io/gha-cache-sunset


pilz:
@ale-rt I will merge this once green to avoid all my PRs to be broken. But please have a look if the caching is still working as intended…

PR #834, opened 2025-04-02
#3384 consultancy

pilz:
Show info on the consultancy view depending on the role of the user.
External users (consultants) don't need to know about the consultancy feature but they need to know if there is work for them to do or not at the moment.

### DIFF
--- a/src/euphorie/client/browser/templates/consultancy.pt
+++ b/src/euphorie/client/browser/templates/consultancy.pt
@@ -130,7 +130,7 @@
           </p>
         </tal:role_consultant>
 
-        <tal:comment tal:replace="nothing">The info about consultancy is only for your own organisation. All external users cannot request it, so they don't need it, and they are likely consultants anyway.</tal:comment>
+        <tal:comment tal:condition="nothing">The info about consultancy is only for your own organisation. All external users cannot request it, so they don't need it, and they are likely consultants anyway.</tal:comment>
         <ul class="link-list"
             tal:condition="not:external_user"
         >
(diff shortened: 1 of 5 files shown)


src/euphorie/client/browser/templates/consultancy.pt
```diff
@@ -33,8 +36,8 @@
           >
             Consultancy
           </h1>
-
-          <p tal:condition="not:role_consultant"
+          <tal:comment tal:replace="nothing">The info that you can request consultancy is only for your own organisation. All external users cannot request it, so they don't need it, and they are likely consultants anyway.</tal:comment>
```
ale-rt, line 39:
```suggestion
          <tal:comment tal:condition="nothing">The info that you can request consultancy is only for your own organisation. All external users cannot request it, so they don't need it, and they are likely consultants anyway.</tal:comment>
```

src/euphorie/client/browser/templates/consultancy.pt
```diff
@@ -118,8 +130,9 @@
           </p>
         </tal:role_consultant>
 
+        <tal:comment tal:replace="nothing">The info about consultancy is only for your own organisation. All external users cannot request it, so they don't need it, and they are likely consultants anyway.</tal:comment>
```
ale-rt, line 133:
```suggestion
        <tal:comment tal:condition="nothing">The info about consultancy is only for your own organisation. All external users cannot request it, so they don't need it, and they are likely consultants anyway.</tal:comment>
```

ale-rt (approved):
Approved, but it is very much better to use `condition:nothing` for comments.

PR #837, opened 2025-04-11
Dsetool: Choices

reinhardt:
Ref https://github.com/syslabcom/scrum/issues/3326

### DIFF
--- a/src/euphorie/client/browser/choice.py
+++ b/src/euphorie/client/browser/choice.py
@@ -25,6 +25,19 @@ class IdentificationView(RiskIdentificationView):
     def selected(self):
         return [option.zodb_path for option in self.context.options]
 
+    @property
+    @memoize
+    def risk(self):
+        # XXX don't inherit this method
+        return self.context.aq_parent.aq_parent.restrictedTraverse(
+            self.context.zodb_path.split("/")
+        )
+
+    @property
+    def is_custom_risk(self):
+        # XXX don't inherit this method
+        return False
+
     def set_answer_data(self, reply):
         answer = reply.get("answer", [])
         if not isinstance(answer, (list, tuple)):
--- a/src/euphorie/client/browser/choice.py
+++ b/src/euphorie/client/browser/choice.py
@@ -1,17 +1,20 @@
-# FIXME fix inheritance
-from euphorie.client.browser.risk import IdentificationView as RiskIdentificationView
 from plone import api
 from plone.memoize.instance import memoize
 from Products.Five import BrowserView
 
 
-class IdentificationView(RiskIdentificationView):
+class IdentificationView(BrowserView):
     """A view for displaying a choice in the identification phase."""
 
     @property
     @memoize
     def webhelpers(self):
         return api.content.get_view("webhelpers", self.context.aq_parent, self.request)
+
+    @property
+    @memoize
+    def navigation(self):
+        return api.content.get_view("navigation", self.context, self.request)
 
     @property
     @memoize
@@ -25,24 +28,16 @@ class IdentificationView(RiskIdentificationView):
     def selected(self):
         return [option.zodb_path for option in self.context.options]
 
-    @property
-    @memoize
-    def risk(self):
-        # XXX don't inherit this method
-        return self.context.aq_parent.aq_parent.restrictedTraverse(
-            self.context.zodb_path.split("/")
-        )
-
-    @property
-    def is_custom_risk(self):
-        # XXX don't inherit this method
-        return False
-
     def set_answer_data(self, reply):
         answer = reply.get("answer", [])
         if not isinstance(answer, (list, tuple)):
             answer = [answer]
         # XXX Check if paths are valid?
+        # for path in answer[:]:
+        #     try:
+        #         self.webhelpers.traversed_session.restrictedTraverse(path)
+        #     except KeyError:
+        #        answer.remove(path)
         self.context.set_options_by_zodb_path(answer)
 
     def __call__(self):
@@ -55,6 +50,5 @@ class IdentificationView(RiskIdentificationView):
         if self.request.method == "POST":
             reply = self.request.form
             self.set_answer_data(reply)
-            return self.proceed_to_next(reply)
-        else:
-            return self.index()
+            return self.navigation.proceed_to_next(reply)
+        return self.index()
--- a/src/euphorie/client/browser/choice.py
+++ b/src/euphorie/client/browser/choice.py
@@ -16,14 +16,20 @@ class IdentificationView(RiskIdentificationView):
     @property
     @memoize
     def choice(self):
-        return self.context.aq_parent.aq_parent.restrictedTraverse(
+        return self.webhelpers.traversed_session.restrictedTraverse(
             self.context.zodb_path.split("/")
         )
+
+    @property
+    @memoize
+    def selected(self):
+        return [option.zodb_path for option in self.context.options]
 
     def set_answer_data(self, reply):
         answer = reply.get("answer", [])
         if not isinstance(answer, (list, tuple)):
             answer = [answer]
+        # XXX Check if paths are valid?
         self.context.set_options_by_zodb_path(answer)
 
     def __call__(self):
--- a/src/euphorie/client/browser/choice.py
+++ b/src/euphorie/client/browser/choice.py
@@ -1,0 +1,41 @@
+# FIXME fix inheritance
+from euphorie.client.browser.risk import IdentificationView as RiskIdentificationView
+from plone import api
+from plone.memoize.instance import memoize
+from Products.Five import BrowserView
+
+
+class IdentificationView(RiskIdentificationView):
+    """A view for displaying a choice in the identification phase."""
+
+    @property
+    @memoize
+    def webhelpers(self):
+        return api.content.get_view("webhelpers", self.context.aq_parent, self.request)
+
+    @property
+    @memoize
+    def choice(self):
+        return self.context.aq_parent.aq_parent.restrictedTraverse(
+            self.context.zodb_path.split("/")
+        )
+
+    def set_answer_data(self, reply):
+        answer = reply.get("answer", [])
+        if not isinstance(answer, (list, tuple)):
+            answer = [answer]
+        self.context.set_options_by_zodb_path(answer)
+
+    def __call__(self):
+        # Render the page only if the user has inspection rights,
+        # otherwise redirect to the start page of the session.
+        if not self.webhelpers.can_inspect_session:
+            return self.request.response.redirect(
+                self.context.aq_parent.absolute_url() + "/@@start"
+            )
+        if self.request.method == "POST":
+            reply = self.request.form
+            self.set_answer_data(reply)
+            return self.proceed_to_next(reply)
+        else:
+            return self.index()
--- a/src/euphorie/client/browser/navigation.py
+++ b/src/euphorie/client/browser/navigation.py
@@ -1,0 +1,82 @@
+from euphorie.client import model
+from euphorie.client.interfaces import CustomRisksModifiedEvent
+from plone import api
+from plone.memoize.instance import memoize
+from Products.Five import BrowserView
+from sqlalchemy import and_
+from z3c.saconfig import Session
+from zope.event import notify
+
+
+class NavigationView(BrowserView):
+    @property
+    @memoize
+    def webhelpers(self):
+        return api.content.get_view("webhelpers", self.context, self.request)
+
+    def proceed_to_next(self, reply):
+        _next = reply.get("next", None)
+        # In Safari browser we get a list
+        if isinstance(_next, list):
+            _next = _next.pop()
+        if _next == "previous":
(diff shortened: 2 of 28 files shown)


src/euphorie/client/navigation.py
```diff
@@ -29,11 +29,29 @@ def FindFirstQuestion(dbsession, filter=None):
...
+    # or an option that is referred to in the condition has been picked
+    # XXX simplify?
+    condition = sql.or_(
+        ~sql.exists().where(model.Choice.id == model.SurveyTreeItem.id),
+        sql.exists().where(
+            sql.and_(
+                model.Choice.id == model.SurveyTreeItem.id,
+                sql.or_(
+                    model.Choice.condition == None,  # noqa: E711
+                    model.Choice.condition.like(model.Option.zodb_path),
```
ale-rt, line 44: This is unclear to me...
You are using `like`, but AFAIK `like` requires a pattern to match.
Is `model.Choice.condition.like(model.Option.zodb_path)` containing an `%`?
If not, it seems to me that this is equivalent to:
```
model.Choice.condition == model.Option.zodb_path
```
If yes, it would be unexpected for me.

src/euphorie/content/browser/choice.py
```diff
@@ -0,0 +1,24 @@
...
+from Acquisition import aq_inner
+from plone.dexterity.browser.view import DefaultView
+from plone.memoize.instance import memoize
+
+
+class ChoiceView(DefaultView, DragDropHelper):
+    @property
+    @memoize
+    def my_context(self):
+        return aq_inner(self.context)
```
ale-rt, line 12: I know this is copied from existing code, but I think this is redundant.

src/euphorie/content/browser/choice.py
```diff
@@ -0,0 +1,24 @@
...
+
+    @property
+    def options(self):
+        return [
+            {
+                "id": option.id,
+                "url": option.absolute_url(),
+                "title": option.title,
+            }
+            for option in self.my_context.values()
```
ale-rt, line 22: What benefit is giving us to decrease the amount of information we have by converting a full object into a `dict`?
It seems to me that returning the `option` instances instead of the dicts would be better.
The advantages would be that you have more information and less code and indirection.

src/euphorie/content/choice.py
```diff
@@ -0,0 +1,84 @@
...
+        "another choice. You can pick multiple options here. If the user selects one "
+        "or more of them, then this choice will be shown. Leave blank to always show "
+        "this choice.",
+        value_type=RelationChoice(
+            vocabulary=StaticCatalogVocabulary({"portal_type": ["euphorie.option"]}),
+        ),
+        required=False,
+        default=[],
+    )
+    directives.widget("condition", SelectFieldWidget)
```
ale-rt, line 46: Is this querying the whole site?
Will the select widget end up having potentially thousands of options to select from?

src/euphorie/upgrade/content/v1/20250402124248_add_content_type__choice/types.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<object name="portal_types">
+  <object meta_type="Conditional Dexterity FTI"
+          name="euphorie.choice"
+  />
+  <object meta_type="Conditional Dexterity FTI"
```
ale-rt, line 6: Does this need to be a `Conditional Dexterity FTI`?
Or is an only addable in choices?

src/euphorie/upgrade/content/v1/20250402124248_add_content_type__choice/types/euphorie.choice.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="utf-8"?>
+<object xmlns:i18n="http://xml.zope.org/namespaces/i18n"
+        meta_type="Dexterity FTI"
```
ale-rt, line 3: Should the meta_type be adapted as well?

src/euphorie/client/model.py
```diff
@@ -147,7 +147,7 @@ class SurveyTreeItem(BaseObject):
         index=True,
     )
     type = schema.Column(
-        Enum(["risk", "module"]),
+        Enum(["risk", "choice", "module"]),
```
ale-rt, line 150: I think that modifying an `Enum` requires an alembic upgrade, but I might be wrong.

ale-rt (commented):
I did a first round of review on the code.
Disclaimer:
- I do not understand much what is going
- I never see the feature live
- I just have a vague idea about what this feature should be

src/euphorie/content/browser/choice.py
reinhardt: Consistency, I guess. It's done the same way with the solutions in the risk view.

src/euphorie/content/choice.py
reinhardt: Yes. That's something on my list. But at least it's possible to select something like this.

src/euphorie/upgrade/content/v1/20250402124248_add_content_type__choice/types.xml
reinhardt: I guess it could also be a regular FTI. Currently there is no restriction.

src/euphorie/client/model.py
reinhardt: I used autogenerate to create the alembic upgrade and it didn't contain anything about the Enum. Also, it works fine as it is.

src/euphorie/content/choice.py
reinhardt: Done.

reinhardt:
Sorry, found another issue. Small fix incoming ...

src/euphorie/client/model.py
ale-rt: Maybe you initialized the DB after the change in Python code was done?
Might I ask you to double check this?

This might help:

- https://stackoverflow.com/questions/14845203/altering-an-enum-field-using-alembic

If after that's not the case and an upgrade is not needed, you can resolve this conversation.

src/euphorie/client/browser/choice.py
```diff
@@ -0,0 +1,107 @@
...
+        # Render the page only if the user can inspection rights,
+        # otherwise redirect to the start page of the session.
+        if not self.webhelpers.can_inspect_session:
+            return self.request.response.redirect(
+                "{session_url}/@@start".format(
+                    session_url=self.webhelpers.traversed_session.absolute_url()
+                )
+            )
+        if self.webhelpers.redirectOnSurveyUpdate():
+            return
```
ale-rt, line 28: I do not understand this if clause.
It seems to me the code is equivalent to:
```suggestion
        self.webhelpers.redirectOnSurveyUpdate():
```
To me

src/euphorie/client/browser/choice.py
```diff
@@ -0,0 +1,107 @@
...
+            if _next == "skip":
+                return self.navigation.proceed_to_next(reply)
+
+            changed = self.set_answer_data(reply)
+
+            if changed:
+                self.session.touch()
+
+            return self.navigation.proceed_to_next(reply)
+        return self.index()
```
ale-rt, line 107: Probably better:
```suggestion
        return super().__call__()
```

src/euphorie/client/browser/navigation.py
```diff
@@ -0,0 +1,105 @@
...
+    def next_question(self):
+        return FindNextQuestion(
+            self.context, dbsession=self.session, filter=self.question_filter
+        )
+
+    def proceed_to_next(self, reply):
+        _next = reply.get("next", None)
+        # In Safari browser we get a list
+        if isinstance(_next, list):
+            _next = _next.pop()
```
ale-rt, line 44: I like more what you did on the other view, where you have a dedicated `_get_next` method.
Any reason to not do the same here?

src/euphorie/client/browser/choice.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def check_render_condition(self):
+        # Render the page only if the user can inspection rights,
+        # otherwise redirect to the start page of the session.
+        if not self.webhelpers.can_inspect_session:
+            return self.request.response.redirect(
+                "{session_url}/@@start".format(
+                    session_url=self.webhelpers.traversed_session.absolute_url()
+                )
+            )
```
ale-rt, line 26: To make the code more readable I would add a `def redirect` method that would make this possible:
```suggestion
            return self.redirect("@@start")
```
or something similar, IDK....  `self.webhelpers.redirect_to_session_path` or whatever.
That can be useful also for the other view.

src/euphorie/client/browser/navigation.py
```diff
@@ -0,0 +1,105 @@
...
+                .filter(
+                    and_(
+                        model.SurveyTreeItem.session == self.session,
+                        model.Module.zodb_path == "custom-risks",
+                    )
+                )
+                .first()
+            )
+            if not sql_module:
+                url = self.context.absolute_url() + "/@@identification"
```
ale-rt, line 79: I am big fan of fstring:
```suggestion
                url = f"{self.context.absolute_url()}/@@identification"
```
They should be even faster, but what you did has nothing wrong.

src/euphorie/client/model.py
```diff
@@ -1486,6 +1507,77 @@ def in_place_custom_measures(self):
...
+        schema.ForeignKey(SurveyTreeItem.id, onupdate="CASCADE", ondelete="CASCADE"),
+        primary_key=True,
+    )
+    condition = schema.Column(types.String(512), nullable=True)
+
+    @property
+    def is_visible(self):
+        if not self.condition:
+            return True
+        options = self.condition.split("|")
```
ale-rt, line 1528: Apparently it is visible only if we have some paths.
What about calling this column `allowed_paths`?

src/euphorie/client/model.py
```diff
@@ -1486,6 +1507,77 @@ def in_place_custom_measures(self):
...
+        if (
+            Session.query(Option)
+            .join(Choice)
+            .filter(Choice.session_id == self.session_id)
+            .filter(Option.zodb_path.in_(options))
+            .count()
+            <= 0
+        ):
+            return False
+        return True
```
ale-rt, line 1538: This is not very readable.
Can count return less than 0?
```suggestion
        return (
            Session.query(Option)
            .join(Choice)
            .filter(Choice.session_id == self.session_id)
            .filter(Option.zodb_path.in_(options))
            .count()
        )
```
but maybe there is also a `.exists()`, I do not remember.

src/euphorie/client/browser/choice.py
reinhardt: Copied from https://github.com/euphorie/Euphorie/blob/781195c2ee117795f3c1d216f94398968a81728b/src/euphorie/client/browser/risk.py#L627

My guess is that it's to make sure that any code that might be added after this line is not executed if the redirect kicks in.

src/euphorie/client/browser/navigation.py
reinhardt: The goal right now is only to move this from `browser/risk.py` so that we can use it for choices, too. Refactoring can be done later.

src/euphorie/client/browser/choice.py
reinhardt: Yes, but let's refactor later and also include `browser/risk.py` to keep things halfways consistent.

src/euphorie/client/model.py
reinhardt: I don't understand. Which column? Why rename it?

src/euphorie/client/model.py
```diff
@@ -1486,6 +1507,77 @@ def in_place_custom_measures(self):
...
+
+    def set_options_by_zodb_path(self, paths):
```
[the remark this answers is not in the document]
reinhardt: I had a hard to getting this to run without errors, I'm reluctant to touch it again.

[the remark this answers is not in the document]
ale-rt: I see your point, would the code simplify if instead of having a relation to an `Option` we would just store the options as a `|` concatenated string?
Then the `option` table could just go away.

src/euphorie/client/model.py
reinhardt: No, sorry, that doesn't fit at all. It's a condition. If any of the options that correspond to the paths is picked, then the choice is visible.

src/euphorie/client/model.py
reinhardt: I haven't initialized this DB in years.

The `type` column is just a varchar, there's nothing to upgrade there.

```
 type            | character varying(6)   |           | not null |
```

src/euphorie/client/model.py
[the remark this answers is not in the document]
ale-rt: Are you making any joins in this PR.
AFAICS the code sort of enforces a 1 to 1 relationship between choices and options.
I might get the code wrong by just reading it.
Let's discuss this in a call.

PR #839, opened 2025-04-25
Fix import of wysiwyg widget.

mauritsvanrees:
Refs https://github.com/syslabcom/scrum/issues/3433

# PRs

- https://github.com/euphorie/NuPlone/pull/60
- https://github.com/euphorie/Euphorie/pull/839
- https://github.com/euphorie/osha.oira/pull/331
- https://github.com/syslabcom/daimler.oira/pull/369
- https://github.com/euphorie/tno.euphorie/pull/22

PR #840, opened 2025-04-27
Directly use the WYSIWYG widget from plonetheme.nuplone 4.0.1.

thet:
Ref: scrum-3433.

This is a new set of PRs to fix the problem with the non-working pat-redactor editor.
This PR used the wysiwyg widgets from plonetheme.nuplone directly and also depends on nuplone >= 4.0.1 where those widgets were introduced resp. copied from plone.app.z3cform (they are not available in recent versions of p.a.z3cform anymore).
I think we can simply do that instead of asking plone.app.z3cform first for their wysiwyg widgets, which are'nt available in Plone 6.1 and should not conflict with our set of wysiwyg widgets in Plone 6.0.

However, I'm asking myself why we're not extending the richtext widgets in plone.app.z3cform to make use of the prepared (but rarly used) extension mechanism for multiple text editors and configure it for pat-redactor..... That should be the cleanest approach, IMO.

## PRs

https://github.com/euphorie/Euphorie/pull/840
https://github.com/euphorie/osha.oira/pull/332
https://github.com/syslabcom/daimler.oira/pull/370
https://github.com/euphorie/tno.euphorie/pull/23

## Obsoletes

https://github.com/euphorie/NuPlone/pull/60
https://github.com/euphorie/Euphorie/pull/839
https://github.com/euphorie/osha.oira/pull/331
https://github.com/syslabcom/daimler.oira/pull/369
https://github.com/euphorie/tno.euphorie/pull/22


ale-rt:
I think this is fine but I would declare that we will support only Plone 6.1, see #841

PR #842, opened 2025-04-29
Change the logo widget

ale-rt:
Do not use anymore the logo widget because the help text is not updated anymore. Update the field help text.

Refs. https://github.com/syslabcom/scrum/issues/3061


pilz (approved):
fine

PR #847, opened 2025-05-02
3454 link checker: truly async, multipass, memory cached

gyst:
Implement a parallellzed, multi-pass, incrementally caching link checker.

Notable changes:
- `requests` is blocking. Use `aiohttp` to get parallel processing
- ~There's a two-layer caching system, that takes care to not memoize connection errors but only store valid status codes in RAM~
- ~This uses a homegrown caching approach using a static, which is kinda hacky. Using a memoizer doesn't work because of reasons documented in the code.~
- This uses a custom approach to memoization, that a) supports asyncio, b) supports the incremental update mechanism we need for this use case. We cannot use the regular memoize decorator, so instead provide our own getter, setter and cache key methods to use against the `plone.memoize.volatile` caching infrastructure.
- Use a multi-pass approach, where the link list is immediately returned, then later incrementally augmented with link status codes
- Incremental passes via `pat-inject` did not work for some reason, I fell back to server-side refreshes instead.
- The multiple passes use an incremental backoff timeout, to combine initial speedy results with ultimately complete status codes
- Some URLs keep being reported as 403 or 404 errors, but show fine in a live browser session. That can be caused by configuration errors, intentional blocking of crawlers, excessive use of Javascript, or whatever. Not something we can fix.

Either the commented-out pat-inject link gets fixed, in which case the server-side refresh needs to be removed; or the pat-inject link should be removed. It's preferable that the injection is fixed, because that would keep scroll position for the user when updates are streaming in; server-side refresh resets the scroll position to top.

### DIFF
--- a/src/euphorie/content/browser/templates/list-links.pt
+++ b/src/euphorie/content/browser/templates/list-links.pt
@@ -39,7 +39,7 @@
 
   <metal:title fill-slot="title">List of Links in ${context/title}</metal:title>
   <metal:content fill-slot="content">
-    <article>
+    <article id="quaive-content">
       <!--a href="${context/absolute_url}/@@list-links?pass=${view/next_pass}"
           class="pat-inject"
           data-pat-inject="trigger: autoload; delay: 0;"


gyst:
Superseded by https://github.com/euphorie/osha.oira/pull/351 and https://github.com/quaive/quaive.osha/pull/163.

Closing this unmerged.

PR #854, opened 2025-05-21
Fix the default language when installing Euphorie

ale-rt:
When installing Euphorie the default language is set to en-us. This language is not listed in the available languages. This can cause the site to return errors when creating new objects that contain the language field.

### DIFF
--- a/src/euphorie/upgrade/deployment/v18/20250521110138_fix_default_language/upgrade.py
+++ b/src/euphorie/upgrade/deployment/v18/20250521110138_fix_default_language/upgrade.py
@@ -10,19 +10,24 @@ class FixDefaultLanguage(UpgradeStep):
         available_languages = api.portal.get_registry_record(
             "plone.available_languages"
         )
-        if default_language not in available_languages:
+        if default_language in available_languages:
+            # No need to change the default language
+            return
 
-            if "en" in available_languages:
-                # Set the default language to English if it's available
-                api.portal.set_registry_record("plone.default_language", "en")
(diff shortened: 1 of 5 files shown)


src/euphorie/upgrade/deployment/v18/20250521110138_fix_default_language/upgrade.py
```diff
@@ -0,0 +1,28 @@
...
+        available_languages = api.portal.get_registry_record(
+            "plone.available_languages"
+        )
+        if default_language not in available_languages:
+
+            if "en" in available_languages:
+                # Set the default language to English if it's available
+                api.portal.set_registry_record("plone.default_language", "en")
+            elif available_languages and len(available_languages) == 1:
+                # Set the default language to the first available language
```
reinhardt, line 19: The comment says “first available language” but the check is `== 1`, so it's the *only* available language. Is that what's intended, or should it be `>= 1`?

reinhardt (commented):
It says “when installing Euphorie”, but there is only an upgrade step, no change to the default profile. Is that intentional?

src/euphorie/upgrade/deployment/v18/20250521110138_fix_default_language/upgrade.py
ale-rt: If we have more than one language, the default language cannot be set with the risk of annoying somebody, for this reason I am raising a `ValueError`.
I will update the comment.

reinhardt (approved):
:sparkles:

PR #857, opened 2025-06-10
Allow the backend users to authenticate on the client

ale-rt:


mauritsvanrees (commented):
I am trying to test this on my Oira Creator copy.  But how is this supposed to work?  I have an account christian_stoney, which I use on both the Quaive and OiRA side.  With that account I can't login on the client, because the client requires that I login with an email address.  When I set an email address on the account, and try to login on the client side with this email, the new code does not find the user.  This would only work when we switch on "use email as login name" in Plone, but I am not sure we want that.

Are you maybe testing with LDAP instead of a local Plone account?

I tried creating a new account for good measure.  This gave errors (see inline comment), but when I worked around those, the login on the client side still did not work, for the same reason.

mauritsvanrees:
Let's assume this works, and a Plone user logs in, and a client account is created.  Later, the user changes his password in Plone.  This currently does not update the password in SQL.   So future client logins will fail.

So we could sync the passwords: run some code whenever the Plone password changes; you may already be planning this.  Or if `self._validate_credentials_on_others(credentials)` passes, we ignore the password and just return the first SQL user that matches in that `authenticate` function.  This needs some moving around of code.

ale-rt:
@mauritsvanrees you should not test this with a "Quoira user" but with a euphorie backend user with an email.

If you are interested in using `allan_neece` you have first to create the user in the Plone `acl_users` of your oira installation.

Indeed, the password sync is a thing, but:
1. I am, not worried about that
2. it is something that might already happen now

ale-rt:
It seems this is more complex to review than what I thought, maybe we can have a call on this next Friday

ale-rt:
I tested freshly in this way

1. I created a user:
```shell
$ ./bin/instance adduser christian_stoney secret
Created user: christian_stoney
```

2: I set up their email:
```shell
(Pdb++) api.user.get_current().setProperties({"email": "christian_stoney@example.com"})
(Pdb++) from transaction import commit
(Pdb++) commit()
```

3. I logged in in the backend.
4. I browsed to the client
5. I am logged in the client

mauritsvanrees:
> I am working on an improvement.

Done: https://github.com/euphorie/Euphorie/pull/858

PR #861, opened 2025-06-23
Activate duplication feature

reinhardt:
Ref syslabcom/scrum#3674

### DIFF
--- a/news/3674.feature
+++ b/news/3674.feature
@@ -1,0 +1,1 @@
+Activate duplication feature if not already active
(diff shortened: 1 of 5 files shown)


news/3674.feature
```diff
@@ -0,0 +1 @@
+Activate duplication feature if not already active
```
ale-rt, line 1: It seems a `breaking` change to me, especially if some customer does not want that.

If this has to be included for everyone, I would just remove the feature flag.

ale-rt (changes requested):
Of course the tests should be fixed

ale-rt (approved):
I would then:

- deprecate the `use_clone_feature`
- make it return always `True`
- remove its usage across the code
- adapt the changelog

Anyway, I approve

news/3674.feature
pilz: changed to breaking.

PR #868, opened 2025-07-14
Allow hiding the report phase until assessment is completed…

reinhardt:
… to a certain percentage

Ref syslabcom/scrum#3237

### DIFF
--- a/news/3237.feature
+++ b/news/3237.feature
@@ -1,0 +1,1 @@
+Allow hiding the report phase until assessment is completed to a certain percentage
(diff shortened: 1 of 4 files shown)


ale-rt (approved):
I approve but please fix the failing pre-commit hook check

reinhardt:
Do you really mean “fix the hook”, or change the template so that the hook passes? I'm not sure how to fix the hook if that's what you mean.

ale-rt:
Of course the second one.
I edited my comment.

reinhardt:
I don't know about the “of course” - to me it's not obvious why there should be a line break and two leading spaces within the sentence. Thanks for clarifying though.

PR #871, opened 2025-07-17
Email reminder: Fix nav entry name

reinhardt:
Needs design, something like

```
#steps>li>a.reminder:not(.onscreen-help):before {
    content:"\e894";
}
```

Ref syslabcom/scrum#3238


angeldasangel:
I added that css snippet to my custom css

PR #872, opened 2025-07-18
Export report_completion_threshold and enable_email_reminder

reinhardt:
Ref syslabcom/scrum#3237 syslabcom/scrum#3238


ale-rt (changes requested):
This is missing a news fragment

PR #875, opened 2025-08-01
Allow the backend users to login to the client

ale-rt:
Allow the backend users to authenticate on the client, provided that they have a valid email

Refs. https://github.com/syslabcom/scrum/issues/3450

### DIFF
--- a/news/+add6b6a1.feature.rst
+++ b/news/+add6b6a1.feature.rst
@@ -1,0 +1,1 @@
+Allow the backend users to authenticate on the client, provided that they have a valid email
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
Just a quick thought.  I wonder if we can easily make SSO work when the backend cookie is valid on the client side as well.  Something like this:

* You are authenticated on the backend.
* You go to the frontend.  You are authenticated (the PAS plugins are the same after all), but you don't have a linked SQL account so you are not authorised and you get redirected to the client login form.
* The GET of the login form sees that you are authenticated already (`api.user.get_current` is an actual user) but are missing a client account.
* We let the login view automatically create a client account and you get redirected to where you came from.

We could try this and see what happens.

PR #883, opened 2025-08-19
Api/complete translations

pilz:
All existing empty message strings translated using AI taking the po file context into account. 
Cross checked translations in german and danish, found good.
Used https://pypi.org/project/gpt-po-translator/


pilz:
Closing, has too many issues. Will create a new one.

PR #894, opened 2025-09-24
Fix survey settings ambiguity

ale-rt:
Since version 19 was released, it was not possible anymore to display the training slides without the questions.

This is due to the fact that the setting "enable_web_training" was used for two purposes: to enable the slides and to enable the questions.

Historically it was only responsible to show the questions.

This patch restores the possibility to display the training slides without the questions by decoupling the two settings. The setting "enable_web_training" enables the slides, while the new "show_training_questions" setting controls the visibility of the questions.

The upgrade step sets "show_training_questions" to True if "enable_web_training" is True and we have training questions.

Refs. https://github.com/syslabcom/scrum/issues/3887


pilz (approved):

ale-rt:
@pilz I added a small fix that I consider an improvement.
If you want to recheck you are welcome.
Otherwise I will merge and deploy to staging once green

PR #895, opened 2025-09-25
Replace Twitter icon with X icon

cillianderoiste:
Refs: https://github.com/syslabcom/scrum/issues/3573


cillianderoiste:
I have't tested this, except to make the equivalent change in the browser inspector to verify that the icon works.

ale-rt (approved):

PR #896, opened 2025-09-26
Test for auto csrf protection with SQLAlchemy.

mauritsvanrees:
This is NOT FOR MERGE!

Refs https://github.com/syslabcom/scrum/issues/3393

* Updates the `@@start` view: if you add `?write=1`, we do a write on GET. Sample url: http://localhost:9660/Plone/client/eu/covid-19/covid-19-revision/++session++79/@@start?write=1 This changes the title of the session, adding a date and time. Extra logging here shows that this makes the SQL session "dirty", so it knows there are changes. This is good.
* Prints debugging info in our custom `EuphorieProtectTransform` to see if we detect changes in SQLAlchemy objects. Result: no, the SQL session here is *not* dirty, so we think there are no changes, so the write on GET succeeds. You can test this in `psql` with `select title from session where id=79;` This is bad.
* Adds a very simple `@@protect-test` page which tries the same thing.  Call this on a country. Sample url: http://localhost:9660/Plone/client/eu/@@protect-test Here it works as we want, and we get redirected to the `@@confirm-action` page. This is good.

I added a few commented-out lines in the `start` view that make a simple change to the Plone site description. Then we *do* get redirected.  So the standard csrf protection still works in this view.

So: theoretically our `EuphorieProtectTransform` transform works, but in practice it does not.

One theory I have is that it would be better in the whole code base to always instantiate an sql Session. So don't do `Session.query(...)` but always `Session().query(...)`. But when I do the "wrong" thing in the protect-test view, it still works.  So as theories go, it seems not so good. ;-)


mauritsvanrees:
I have found it.  The problem is flushing.

* Edit an SQLAlchemy item. `session.dirty` now contains the item.
* Do a query.  This does a flush, because `z3c.saconfig` has determined that this is the [best for Zope integration](https://github.com/zopefoundation/z3c.saconfig/blob/2.0/src/z3c/saconfig/utility.py#L45-L51).
* Now `session.dirty` is empty again.

This is shown in commit 6951ff7ac948de05bfb83fe6aac7b49783c81980.

One way to work around this, is to set `autoflush=False`.  See commented-out code in commit 815a4cdaa33c7666c2fcfc79a702da3e51af6826.  That helps.

Problem with that: even if we would not use auto flush, there may be valid reasons to _explicitly_ flush the session.  We may create a row in a table, and then want to query the table so we can know what the row id is.  This needs a flush.  If this is done in a GET request, the auto csrf protection should still kick in.

BTW, note that a flush is still different from a commit: I have made changes to a row, then called a query with auto flush, then in another tab with `psql` checked the row, and this did not yet have the change.

Solution: in our `plone.protect` override, do an additional check for `resource.tx._dirty`, next to `resource.session.dirty`.  See last commit 6fcb4964882ee560e7d3f3347ee34884076be119.  Yes, this accesses an internal attribute, but I see no other way.

@ale-rt Do you agree we should go with that last commit?  I would create a new PR with only the changes in `protect.py`, but cleaned up.  And hopefully I can add some tests.

ale-rt:
That's a brilliant summary :)
Feel free to go with a new PR or whatever you like.

:clap:

mauritsvanrees:
I close this.  Continued in https://github.com/euphorie/Euphorie/pull/897.

PR #922, opened 2026-03-06
Switch to collective.ftw.upgrade

ale-rt:


setup.py
```diff
@@ -37,6 +37,7 @@
         "alembic",
         "anytree",
         "Chameleon",
+        "collective.ftw.upgrade",
         "decorator",
         "ftw.upgrade",
```
mauritsvanrees, line 42: I suppose `ftw.upgrade` could be removed from `setup.py`.

mauritsvanrees (approved):

setup.py
ale-rt: I will remove it in another PR

PR #930, opened 2026-04-27
Fix the sidebar markup to match proto

ale-rt:
Refs. https://github.com/syslabcom/scrum/issues/4754

### DIFF
--- a/src/euphorie/client/browser/templates/shell.pt
+++ b/src/euphorie/client/browser/templates/shell.pt
@@ -236,15 +236,15 @@
                   <li class="${data/class}"
                       id="${data/id}"
                   >
-                    <a class="${data/name} pat-switch pat-inject"
+                    <a class="${data/name} pat-switch pat-inject ${python:'step' if data['id'].startswith('step-') else ''}"
                        href="${data/href}"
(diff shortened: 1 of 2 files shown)


src/euphorie/client/browser/templates/shell.pt
```diff
@@ -258,7 +258,9 @@
                      data-pat-inject="history: record"
                      data-pat-switch="selector: #main; remove: sidebar-*; add: sidebar-off"
                      i18n:translate="label_exit"
-                  >Exit</a>
+                  >
+                    <span class="label">Exit</span>
```
mauritsvanrees: The `i18n:translate="label_exit"` should be moved from the a-tag to the span.

(I wanted to add a suggestion, but that does not seem to work, probably because it would need to touch lines that have not been changed.)

mauritsvanrees (changes requested):
That looks much better, thanks.
One tiny change needed for i18n, for the rest this is fine.

ale-rt:
I pushed a fixup commit, I will merge once it is is green

mauritsvanrees (approved):

PR #931, opened 2026-04-28
Update the markup to fix the edit slide icon

ale-rt:
Refs. https://github.com/syslabcom/scrum/issues/4718

### DIFF
--- a/news/4718.bugfix.rst
+++ b/news/4718.bugfix.rst
@@ -1,0 +1,1 @@
+Update the markup to fix the edit slide icon. @ale-rt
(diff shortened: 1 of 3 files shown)


ale-rt:
See left (staging) and right (dev with this patch):
image

ale-rt:
I pushed another small fix that changes the injection parameters to match the ones in proto for the sidebar links:
image
Before it was reinjecting the whole `#content` giving a flashy effect.

mauritsvanrees (approved):
The new markup is good.

I'm not sure about the pat-inject change.  I don't see a problem without it.

What I do see, both with and without that change:

* Within the training I click to edit a question.
* I tick or untick a measure and click Save.
* The page shows no change: the measure is still shown or hidden as before.
* Then I reload the page and the new status is correctly visible.

So that may need a change still.

But the icon problem is fixed, and I see no bad side effects of the pat-inject change, so I approve.

ale-rt:
The second commit fixes a "flash" visible when switching through the various steps in the sidebar.

PR #932, opened 2026-04-29
Fix typo when reporting all roles for country managers.

mauritsvanrees:
This had ``CountryyManager`` with two instead of one ``y``.

There were two tuples, with this typo as only difference. So define the tuple once.

Refs https://github.com/syslabcom/scrum/issues/4282


ale-rt (approved):
You fixed a 14 years old bug 🤣

PR #934, opened 2026-04-30
Update translations 2026-04-30

mauritsvanrees:
This is for https://github.com/syslabcom/scrum/issues/4764

It uses the updated script from https://github.com/euphorie/euphorie-project/pull/48

There are very many changes, making this hard to review, but it can be useful to look at the individual commits.


ale-rt (approved):

pilz:
I checked some translation differences with angel and found that we have to do some more cleanup first. Angel is preparing a minimal PR with only her 5 translations at the moment. Afterwards, I will sort out the translations with her and we will update this PR once more.

pilz:
I looked deeper into that idea of taking over translations from proto and while it is a noble idea, we are not ready for it yet. Proto has again evolved in such a way that taking over the translations would introduce 15 translations which are genuinely wrong.

Not taking over the translations from proto means that the help files differ from reality, but that is an order of magnitude smaller a problem than writing wrong statements into dev.

I retract the idea to update translations automatically from proto until further notice.
We will amend translations manually where we see fit.

PR #940, opened 2026-05-08
Do not compile HTML static resources

ale-rt:
Reources like:
```
++resource++euphorie.resources/assets/oira/help/en/illustrations/report-download.html
```
can throw an error if they because Chameleon tries to compile them.

Refs. https://github.com/syslabcom/scrum/issues/4798

### DIFF
--- a/news/4798.feature.rst
+++ b/news/4798.feature.rst
@@ -1,0 +1,1 @@
+Fix errors that might happen in help pages when the template engine tries to compile them. @ale-rt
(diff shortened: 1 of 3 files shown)


ale-rt:
The tests should be fixed by: https://github.com/euphorie/Euphorie/pull/941

thet (approved):
👍 
Very interesting solution to that problem, I like it!

mauritsvanrees (approved):
I don't see any missing help pages without this.  But the change makes sense.

PR #942, opened 2026-05-12
Remove the unused xmlimport script

ale-rt:
It has broken imports since 2020 when the publish module was moved to the browser folder.

Refs. https://github.com/syslabcom/scrum/issues/4776


mauritsvanrees (approved):
LGTM

Maybe the `init-db` command should be removed as well.  At least the instruction in `install.rst` does not work for me:

```
$ bin/instance init-db
*** Unknown syntax: init-db
```

I guess `zopectl.command` is no longer supported, or has changed.

ale-rt:
Uhm, I might have a look but this is low prio

PR #947, opened 2026-06-05
Add ``readonly`` parameter to ``CopyToClient``.

mauritsvanrees:
Useful for getting an existing client survey or preview. This is for https://github.com/syslabcom/scrum/issues/4608

### DIFF
--- a/src/euphorie/client/browser/publish.py
+++ b/src/euphorie/client/browser/publish.py
@@ -31,12 +31,14 @@ import secrets
 log = logging.getLogger(__name__)
 
 
-def CopyToClient(survey, preview=False):
+def CopyToClient(survey, preview=False, readonly=False):
     """Copy the survey to the online client part of the site.
 
     :param survey: the survey to copy
     :param bool preview: indicates if this is a preview or a normal publication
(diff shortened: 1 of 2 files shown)


src/euphorie/client/browser/publish.py
```diff
@@ -81,17 +85,22 @@ def CopyToClient(survey, preview=False):
...
+        target.logo = sector.logo
+        # Clear any scaled logos
+        AnnotationStorage(target).storage.clear()
 
     if preview:
         new_id = "preview"
     else:
         new_id = surveygroup.id
+    if readonly:
+        return getattr(target, new_id, None)
```
ale-rt, line 103: Is this acquisition safe?

ale-rt (approved):
LGTM, it might be worth do check if the method is acquisition safe

src/euphorie/client/browser/publish.py
mauritsvanrees: What do you mean with acquisition safe?  It works for me, I get an answer back.

ale-rt: I mean that we could have a new_id upper in the acquisition chain.
Nevermind.

PR #950, opened 2026-06-16
Fix DOCX HTML conversion for text-only markup

ale-rt:
Refs. https://github.com/syslabcom/scrum/issues/4885

### DIFF
--- a/src/euphorie/client/docx/html.py
+++ b/src/euphorie/client/docx/html.py
@@ -116,19 +116,31 @@ class HtmlToWord:
     def __call__(self, markup, doc, style=None, next_style=None):
         if not markup or not markup.strip():
             return doc
+
         try:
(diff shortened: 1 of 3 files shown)


src/euphorie/client/docx/html.py
```diff
@@ -116,19 +116,31 @@ def handleElement(self, node, doc, style=None):
...
         try:
             markup_doc = lxml.html.document_fromstring(markup)
         except etree.XMLSyntaxError:
             text = strip_markup(markup)
             text = text.replace("&#13", "\n")
-            doc.add_paragraph(text)
+            doc.add_paragraph(text, style=style)
             return doc
 
-        for idx, node in enumerate(markup_doc.find("body")):
```
ale-rt, line 127: The bug comes from the fact that when we have some plain text the `body` element has no children, just a `text` attribute.

PR #953, opened 2026-06-30
Allow to customize the static resources used in the client

ale-rt:
Refs. https://github.com/syslabcom/scrum/issues/4926

### DIFF
--- a/src/euphorie/client/browser/webhelpers.py
+++ b/src/euphorie/client/browser/webhelpers.py
@@ -90,21 +90,8 @@ class WebHelpers(BrowserView):
     View name: @@webhelpers
     """
 
-    certificates_path = "++resource++euphorie.resources/assets/oira/certificates"
-    media_path = "++resource++euphorie.resources/media"
-    style_path = "++resource++euphorie.resources/assets/oira/style"
     script_path = "++resource++patternslib"
-
-    brand = "oira"
-
-    css_path = "++resource++euphorie.resources/assets/{brand}/style/all.css"
-    css_path_min = "++resource++euphorie.resources/assets/{brand}/style/all.css"
-
     js_name = "bundle.min.js"
-
-    favicon_path = (
-        "++resource++euphorie.resources/assets/{brand}/favicon/apple-touch-icon.png"
-    )
 
     group_model = Group
     hide_organisation_tab = False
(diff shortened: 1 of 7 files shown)


src/euphorie/client/browser/webhelpers.py
```diff
@@ -126,9 +113,68 @@ def to_decimal(self, value):
...
+
+    @property
+    @memoize_contextless
+    def certificates_path(self) -> str:
+        return f"{self.resources_base_path}/certificates"
+
+    @property
+    @memoize_contextless
+    def css_path(self) -> str:
+        return f"{self.resources_base_path}/style/all.css"
```
mauritsvanrees:
```suggestion
        return f"{self.style_path}/all.css"
```

src/euphorie/client/browser/webhelpers.py
```diff
@@ -126,9 +113,68 @@ def to_decimal(self, value):
...
+
+    @property
+    @memoize_contextless
+    def css_path_min(self) -> str:
+        return f"{self.resources_base_path}/style/all.css"
```
mauritsvanrees:
```suggestion
        return f"{self.style_path}/all.css"
```

src/euphorie/client/browser/webhelpers.py
```diff
@@ -623,39 +678,25 @@ def media_url(self):
...
     @memoize
     def js_url(self):
-        return "{}/{}/{}?t={}".format(
-            self.client_url,
-            self.script_path,
-            self.js_name,
-            self.resources_timestamp,
-        )
+        script_base_url = f"{self.client_url}/{self.script_path}"
+        return f"{script_base_url}/{self.js_name}?t={self.resources_timestamp}"
```
mauritsvanrees:
```suggestion
        qs = urlencode({"t": self.resources_timestamp})
        return f"{script_base_url}/{self.js_name}?{qs}"
```

src/euphorie/client/browser/webhelpers.py
```diff
@@ -623,39 +678,25 @@ def media_url(self):
     @property
     @memoize
     def style_url(self):
-        return f"{self.client_url}/{self.style_path}"
+        return f"{self.resources_base_url}/style"
```
mauritsvanrees: I like the previous code better.  Now if you change the `style_path` you need to manually update `style_url` as well.  Same for a few other places where you hardcode `/style` or `all.css`.

Alternatively, maybe we should deprecate some of these `*_path` properties, since the `*_url` properties seem to be the only ones that are used by templates that use this browser view.  Then a bit of hardcoding/duplication is okay.

src/euphorie/client/tests/test_webhelpers.py
```diff
@@ -740,6 +741,101 @@ def test_survey_tree_data__consultancy(self):
                             self.assertEqual(nav_tree[4]["disabled"], False)
                             self.assertEqual(nav_tree[5]["disabled"], False)
 
+    def test_resource_urls(self):
```
mauritsvanrees: I would prefer splitting this test in two.  Something like `test_resource_urls_default` and `test_resource_urls_custom`.

mauritsvanrees (changes requested):
The approach is good, but I have some suggestions.

The [tests fail](https://github.com/euphorie/Euphorie/actions/runs/28453094620/job/84320399896?pr=953#step:8:73) with:

```
KeyError: 'Duplicate profile ID: euphorie.content:default'
```

Which is weird at first, but it is caused by this:

```
ModuleNotFoundError: No module named 'collective.ftw'
```

So you should add `collective.ftw.upgrade` to the installation requirements.  Or maybe first in a separate PR replace all calls of `ftw.upgrade` to `collective.ftw.upgrade`.

ale-rt:
> So you should add `collective.ftw.upgrade` to the installation requirements. Or maybe first in a separate PR replace all calls of `ftw.upgrade` to `collective.ftw.upgrade`.

Yes there is already a PR for that that I need to finalize:
- https://github.com/euphorie/Euphorie/pull/922/

src/euphorie/client/browser/webhelpers.py
ale-rt: I think there are way to many of this properties.
I agree with the deprecation path but I wanted to separate it from this task.
Let me think whether to include it or not.

ale-rt:
I rebased this branch over https://github.com/euphorie/Euphorie/pull/922 
I think the fixes you requested have been properly applied.

mauritsvanrees (approved):
LGTM now, thanks.

